PR #218, opened 2013-06-09
SQLite adapter: delete only if table has been used

ghost:
The SQLite adapter lacks the functionality to only truncate those tables which were used. I modified this adapter to check whether tables were used before the truncation is executed.
I don't know whether this functionality was not yet implemented on purpose, but for me this gave speed-up in some cases. It only works if the table `sqlite_sequence` is used by SQLite.

From the [SQLite documentation](http://www.sqlite.org/fileformat2.html#seqtab):
- Table has a primary key. Then when the table is modified, the entry in `sqlite_sequence` is created/modified, so there is an entry in the `sqlite_sequence` table. If this entry exists, it must be removed and the table must be cleared if rowcount > 0.
- Table has no primary key, then there is no auto increment value to be reset, so the table must be cleared exactly if rowcount > 0.

Note: the SQLite adapter (still) ignores the configuration options `reset_ids` and `pre_count`.

### DIFF
--- a/lib/database_cleaner/active_record/truncation.rb
+++ b/lib/database_cleaner/active_record/truncation.rb
@@ -93,14 +93,23 @@ module DatabaseCleaner
 
     module SQLiteAdapter
       def delete_table(table_name)
-        execute("DELETE FROM #{quote_table_name(table_name)};")
         if uses_sequence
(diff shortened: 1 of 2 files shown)


botandrose:
Thank you for the PR! However, the ActiveRecord adapter has been extracted into its own repo at https://github.com/DatabaseCleaner/database_cleaner-active_record. If you close this PR and reopen it over there, I'd be happy to take a look!

Issue #309, opened 2014-12-13
problem to clean database when change config


wendy0402:
Hi, previously I got a problem here. so before I tried to configure the database cleaner in my rspec(I am using activerecord).
the code as below:

``` ruby
DatabaseCleaner[:active_record, connection: :local].strategy = :truncation
DatabaseCleaner[:active_record, connection: :mysql].strategy = :truncation
```

but it doesn't work. and after I trace it, I found out my problem. It is because I change my db config in the run time and when database cleaner configuration using connection instead of model, by default it tried to search configuration from `db/config.yml`. I already fixed my problem by using models

but I'm curious why it doesn't use `::ActiveRecord::Base.configurations`  to get the database configuration ? it get the same configuration with `db/config.yml` and it will be more reliable to handle configuration changes in runtime.

PR #310, opened 2014-12-13
use ActiveRecord::Base configuration when different with config file

wendy0402:
this is related with my issues #309 . So I tried to create this solution. when it different with config file, will use the ActiveRecord::Base configuration instead. Please review it, I am still newbie so maybe the code is not good enough


etagwerker:
@wendy0402 Great, thanks!

PR #311, opened 2014-12-19
Adding 'ensure' to 'cleaning' method

cema-sp:
Adding 'ensure' to 'cleaning' method for better error tolerance


etagwerker:
@cema-sp I like this. Could you add an rspec scenario that shows how to this improves the code?

etagwerker:
@cema-sp Thank you! :smile:

PR #314, opened 2014-12-31
clears neo4j indexes and constraints

subvertallchris:
Fully cleaning out a Neo4j database also requires dropping indexes and constraints. We haven't added support for querying the endpoints that provide the complete index/constraint lists in the neo4j-core gem, though we should add it into the next version. In the meantime, this uses Net/HTTP to perform the query, then we do the rest through Cypher.

### DIFF
--- a/lib/database_cleaner/neo4j/deletion.rb
+++ b/lib/database_cleaner/neo4j/deletion.rb
@@ -7,9 +7,33 @@ module DatabaseCleaner
       include ::DatabaseCleaner::Neo4j::Base
 
       def clean
-        ::Neo4j::Transaction.run do
-          session._query('MATCH (n) OPTIONAL MATCH (n)-[r]-() DELETE n,r')
+        ::Neo4j::Transaction.run { session._query('MATCH (n) OPTIONAL MATCH (n)-[r]-() DELETE n,r') }
+        drop_indexes_and_constraints
+      end
+
+      private
+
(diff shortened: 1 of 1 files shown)


etagwerker:
@subvertallchris Thank you for this pull request! 

Has the support for index/constraint lists changed in neo4j-core? I'm not familiar with neo4j but I found this: https://github.com/neo4jrb/neo4j-core/wiki/Label-Index#index-and-constraint-maintenance 

Also, please add a spec for the new behavior.

subvertallchris:
Yes, I added those methods a few months ago because this process really deserved to be part of the gem. I'll add specs as soon as I can. Had totally forgotten I sent this over!

etagwerker:
@subvertallchris Sounds good. Thank you!

stringsn88keys:
@subvertallchris Have you had an opportunity to look into adding specs for PR?

cheerfulstoic:
I know that @subvertallchris hasn't been working on Neo4j lately, but I'm actually not sure that we always want to clear indexes and constraints in Neo4j.  Apps will generally have one set of indexes / constraints (just like in a relational database) and they won't change.

@subvertallchris might be thinking about our tests in the `neo4j` and `neo4j-core` gems where we do need to clear out the schema elements because we're testing different situations that the gems will need to deal with.

botandrose:

PR #315, opened 2015-01-02
Removed jeweler from the gem.

etagwerker:
Hey @DatabaseCleaner/owners 

I don't think `jeweler` is necessary. I suggest that we get rid of it and use `gem` instead.

What do you think? 

Ernesto


JonRowe:
I propose we use bundlers tasks, they're very easy to use and compatible with these changes.

PR #319, opened 2015-01-05
Run tests against the latest versions of Ruby 2.1 and 2.2 on travis

deepj:


deepj:
Probably this will require use more recent version of `mysql` (`2.9.1` perhaps?) for Ruby 2.2. I don't use `mysql` at all so I can't try it.

vrinek:
Incorporated in https://github.com/DatabaseCleaner/database_cleaner/pull/329

PR #316, opened 2015-01-05
Remove stray require 'mongoid/tree'

TimMoore:
Closes #299.

PR #320, opened 2015-01-05
Fix `cannot load such file -- mongoid/tree (LoadError)` (Mongoid 4.0.0)

mkempe:
Hey there,

i just updated to version `1.4.0` and with Mongoid `4.0.0` an error `cannot load such file -- mongoid/tree (LoadError)` occurred.

I just looked at the history, with the commit https://github.com/DatabaseCleaner/database_cleaner/commit/61feda02e0999a684951f2afb6746fd2ddf814f4 the line `require 'mongoid/tree'` was added. Is this needed, without the line everything works?

PS: Maybe @etagwerker should take a look at it.

PPS: Many thanks for database_cleaner!


mkempe:
Just saw that #316 already exist. Sorry for the noise.

PR #328, opened 2015-01-12
except tables not working with postgresql schema prefix

phoenixek12:
I define tables as strings containing only name of table and in postgres i get schema prefix in name. So simple array - not working for me. With this change except work on postges. Its a bit slower, but it keeps schema prefix and works as expected.

### DIFF
--- a/lib/database_cleaner/active_record/truncation.rb
+++ b/lib/database_cleaner/active_record/truncation.rb
@@ -244,8 +244,8 @@ module DatabaseCleaner::ActiveRecord
       tables_in_db = cache_tables? ? connection.database_cleaner_table_cache : connection.tables
       to_reject = (@tables_to_exclude + connection.database_cleaner_view_cache)
       (@only || tables_in_db).reject do |table|
-          if table.match(/([^.]+)$/)
-            to_reject.include?($1)
+          if ( m = table.match(/([^.]+)$/) )
+            to_reject.include?(m[1])
           else
             false
           end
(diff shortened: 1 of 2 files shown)


lib/database_cleaner/active_record/truncation.rb
```diff
@@ -242,7 +242,14 @@ def clean
 
     def tables_to_truncate(connection)
       tables_in_db = cache_tables? ? connection.database_cleaner_table_cache : connection.tables
-      (@only || tables_in_db) - @tables_to_exclude - connection.database_cleaner_view_cache
+      to_reject = (@tables_to_exclude + connection.database_cleaner_view_cache)
+      (@only || tables_in_db).reject do |table|
+          if table.match(/([^.]+)$/)
+            to_reject.include?($1)
```
vrinek: Please do not use `$1`. These types of variables are too cryptic for most developers and there are always ways around them. In this case, please assign the result from `table.match(/([^.]+)$/)` to a local variable and use that one.

phoenixek12: Ok, I did it that way. Is it ok now?

phoenixek12:
@vrinek Hi, what about this pull request? :(

vrinek:
I'm sorry, I haven't found time to test it. I'll try to test it soon and merge it.

etagwerker:
@phoenixek12 Could you add a scenario that shows the problem you're seeing?

phoenixek12:
@etagwerker Is this good enough?

etagwerker:
@phoenixek12 Yes, thanks!

Issue #317, opened 2015-01-05
v1.4.0 cleans the 'schema_migrations'


nepalez:
It seems, in v1.4.0 the cleaner cleans the 'schema_migrations' table as well.

This causes the next db migration to break, because the migration task tries to duplicate old migrations from the very beginning.

geoffharcourt:
I'm seeing the same behavior in my app in Rails 4.1.

enthrops:
Confirming on Rails 4.2

bdarfler:
Confirmed on Postgres/ActiveRecord/Rails 4.1. It seems that the logic now requires the full `database.table` notation to work properly. A work around would be to add `except: %w[public.schema_migrations]` but obviously this should be fixed. It also impacts anyone trying to prevent `spatial_ref_sys` from being truncated.

uno4ki:
Yeah, nasty bug.
Confirmed on 4.2 with postgres, ```public.*``` is a lifesaver

raphaelcosta:
Same here with Rails 4.2

felipero:
+1

lilfaf:
:+1:

jonallured:
I recently upgraded to Rails 4.2 and mistakenly thought that had something to do with my `schema_migrations` table was getting truncated after cucumber runs. I eventually found this issue and I tried the work around of using the `except` option, but that didn't work.

I ended up downgrading to v1.3.0 for the time being, but I also cloned down the project and ran the tests. Everything is passing and there is a spec for this behavior:

https://github.com/DatabaseCleaner/database_cleaner/blob/master/spec/database_cleaner/active_record/truncation_spec.rb#L36

Any thoughts on how to get a failing test on this? I couldn't help notice that the library runs against `activerecord` 3.0.0, so that can't be good...

Thanks!
Jon

jonallured:
I'm new to the internals of this library, so I spent some time reading through things this afternoon and something that stood out to me was the lack of outer loop testing for truncation and the `schema_migrations` table. What I was looking for was a test that migrates the database, truncates and then asserts about the state of the `schema_migrations` table, but I couldn't find a test like that. If I've missed something, please let me know!

I took a stab at adding one - both on top of v1.4.0 and off of v1.3.0 in order to ensure that the test was actually proving something. Here are the commits:

* jonallured/database_cleaner@04c263e - passing spec off of v1.3.0
[...]

PR #337, opened 2015-02-04
Exclude schema_migrations from tables_with_schema

palkan:
This is a fix for https://github.com/DatabaseCleaner/database_cleaner/issues/317 .
The easiest way to handle the problem (as I think) is to exclude migrations table from table list while loading it, because we don't know anything about schema. Thus generic `migration_storage_names` makes no sense when using with Postrges.


robpark:
:+1:

gkop:
+1

etagwerker:
@palkan Great, thanks!

PR #340, opened 2015-02-10
removed dependency of Mongo adapter to Mongoid

codegourmet:
I'm not happy with including/requiring Mongoid ORM if I'm on pure mongo.rb database adapter.


etagwerker:
@codegourmet Ok, seems reasonable. Thanks!

PR #343, opened 2015-03-06
Fix truncation on MongoDB 3

andreale:
As reported on the [mongoDB documentation](system.namespaces), the collection `system.namespaces` is deprecated and can no longer be used to list the database collections.

5ca89421407c3a033eed20b4997a64c6

This PR addresses the deprecation issue by using the command `listCollections` instead. This allow DatabaseCleaner to work with MongoDB 3 with WiredTiger storage engine.

### DIFF
--- a/lib/database_cleaner/moped/truncation_base.rb
+++ b/lib/database_cleaner/moped/truncation_base.rb
@@ -23,10 +23,11 @@ module DatabaseCleaner
           session.use(db)
         end
 
-        session.command(listCollections: 1)[:cursor][:firstBatch].map do |collection|
-          collection[:name]
-        end.reject do |collection_name|
-          collection_name =~ /system|\$/
+        session.command(listCollections: 1)['cursor']['firstBatch'].map do |collection|
+          collection['name']
+        end
+        .reject do |collection_name|
+          collection_name =~ /\.system\.|\$/
         end
       end
 
--- a/lib/database_cleaner/moped/truncation_base.rb
+++ b/lib/database_cleaner/moped/truncation_base.rb
@@ -23,11 +23,18 @@ module DatabaseCleaner
           session.use(db)
         end
 
-        session.command(listCollections: 1)['cursor']['firstBatch'].map do |collection|
-          collection['name']
-        end
-        .reject do |collection_name|
-          collection_name =~ /\.system\.|\$/
+        if db_version.split('.').first.to_i >= 3
+          session.command(listCollections: 1)['cursor']['firstBatch'].map do |collection|
+            collection['name']
+          end
+          .reject do |collection_name|
+            collection_name =~ /\.system\.|\$/
+          end
+        else
+          session['system.namespaces'].find(name: { '$not' => /\.system\.|\$/ }).to_a.map do |collection|
+            _, name = collection['name'].split('.', 2)
+            name
+          end
         end
       end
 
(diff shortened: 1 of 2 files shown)


zjx20:
Thanks! This really helps a whole lot!

bejmuller:
+1

FlammenB:
:+1:

lifegiver:
:ok:

simonmorley:
+1

saschaknobloch:
:+1:

lib/database_cleaner/moped/truncation_base.rb
```diff
@@ -23,9 +23,18 @@ def collections
...
 
-        session['system.namespaces'].find(:name => { '$not' => /\.system\.|\$/ }).to_a.map do |collection|
-          _, name = collection['name'].split('.', 2)
-          name
+        if db_version.split('.').first.to_i >= 3
+          session.command(listCollections: 1)['cursor']['firstBatch'].map do |collection|
+            collection['name']
+          end
+          .reject do |collection_name|
+            collection_name =~ /\.system\.|\$/
```
saschaknobloch: This regexp isn't matching `system` collections. It should be `collection_name =~ /.?system\.|\$/`

For MongoDB 3.x.x using Wired Tiger storage engine this PR is sufficient, because by default there are not `.system.` collections. But if you use MongoDB 3.x.x with MMapV1 storage engine, it will fail, because there still is the `system.indexes` collection.

andreale:
Hi @saschaknobloch, thanks for your feedback.

I just reused the (already present) regexp that filters out any `xxxxxx.system.indexes` and `xxxxxx.yyyyyy.$_id_` collection. This should make my change also compatible with MMAPv1 collections (by excluding those collections form the truncation).

Could you please clarify better the issue?

saschaknobloch:
Hey @andreale,

here is a link to rubular with the regexp provided with this code, check out yourself:

http://rubular.com/r/qpc2cGb38J

saschaknobloch:
BTW: the command `listCollections` allows also for setting a `filter` 

http://docs.mongodb.org/manual/reference/command/listCollections/#dbcmd.listCollections

This gives us the chance to solve this problem more elegant (but still the regex provided with this PR doesn't work for me ;) ):

```
session.command(listCollections: 1, filter: {"name"=> {"$not"=>/.?system\.|\$/}})['cursor']['firstBatch'].map do |collection| 
...
end 
```

andreale:
Thanks @saschaknobloch.

I could reproduce the issue on another dev machine (where we also get a `<database_name.$cmd` collection) and confirm your updated regexp is working. I'm updating my PR, with also the filter at mongo-level.

- Ale

lucke84:
+1

dot:
+1

saschaknobloch:
Since there is no progress on supporting MongoDB 3, we've released our own implementation: 

https://github.com/td-berlin/mongoid_cleaner

We've added `drop` strategy, which improves the performance (compared to `truncation` strategy) by factor 3-5 for our projects.

It only supports Mongoid and MongoDB 3. It's tested with:

- Ruby 2.1
- Ruby 2.2
- JRuby 
- Ruby-Head

Each against WiredTiger and MMAPv1. 

Feel free to use and contribute.

etagwerker:
@andreale Thank you!

PR #344, opened 2015-03-11
fix indent in active_record/truncation.rb

HaiTo:
fix indent and unnecessary space in `truncation.rb`


etagwerker:
@HaiTo Thanks!

PR #333, opened 2015-01-21
#332 - mysql2+jruby support, speedup

ahorek:
#332

PR #352, opened 2015-03-25
Fixed deletion strategy for JDBC MySQL adapter

DanElbert:
It seems the JDBC MySQL adapter creates an ActiveRecord::ConnectionAdapters::Mysql2Adapter constant that causes the SelectiveTruncation module to be included in the active record Deletion class.  Everything works fine except the real MySQL adapter returns results differently from the JDBC adapter.  Changing the execute method to exec_query allows it to work for both adapters.


ahorek:
https://github.com/DatabaseCleaner/database_cleaner/pull/333

etagwerker:
@DanElbert Thank you! :smile_cat:

PR #353, opened 2015-04-07
Ensure @cleaners has been defined before reading

orien:
To avoid Ruby warnings like:
```
.bundle/ruby/2.1.0/gems/database_cleaner-1.4.1/lib/database_cleaner/configuration.rb:45: warning: instance variable @cleaners not initialized
```


etagwerker:
@orien Awesome, thanks!

PR #356, opened 2015-04-20
Reduce mongoid_collection_names complexity

mrb:
I was browsing Code Climate for some OSS repos to contribute to, and saw that database_cleaner was already on Code Climate with an excellent GPA! Not much to change here but I figured I'd help clarify a method that Code Climate flagged as complex. I also added our badge to your README. Cheers!


etagwerker:
@mrb Cool, thanks!

Issue #273, opened 2014-06-01
truncation in capybara tests results in rspec freezing (while deletion works)


steookk:
I found out that this code (necessary to truncate in case we are testing with capybara) makes RSpec freeze right after running capybara tests, while the same code with :deletion strategy works smoothly. 
```ruby
    config.before(:suite) do
      DatabaseCleaner.strategy = :transaction
    end

    config.around(:each) do |example|
      if example.metadata[:type] == :feature
        example.run
        DatabaseCleaner.clean_with :truncation #ONLY DELETION WORKS!
      else 
        DatabaseCleaner.start
        example.run
        DatabaseCleaner.clean
      end
    end
```
As a matter of fact, the following code works for truncation, which makes me think that DatabaseCleaner.start must be run for truncation as well.
```ruby 
    config.before(:each) do
      DatabaseCleaner.strategy = :transaction
    end
    config.before(:each, :type => :feature) do
      DatabaseCleaner.strategy = :truncation
    end
    config.before(:each) do
      DatabaseCleaner.start #runs for both strategies
    end

    config.after(:each) do
      DatabaseCleaner.clean
    end
````
I don't know whether this is a bug or a some kinds of choice due to some reasons, however I felt it was worth pointing it out.

**Comment by karpah** [comment]
Unfortunately nothing here has fixed the problem in my app.

The test running sits there doing nothing, and when I start a Postgres console and check the `pg_stat_activity` table, I see the two connections sitting there, one active and one idle in transaction doing nothing. I don't know what's going on.

**Comment by lucashungaro** [comment]
@karpah Can you paste your spec_helper here?

**Comment by airatshigapov** [comment]
I am experiencing the same issue:

This config causes specs to freeze:

```ruby
RSpec.configure do |config|
  config.before(:suite) do
    DatabaseCleaner.strategy = :transaction
    DatabaseCleaner.clean_with(:truncation)
  end

  config.around(:each) do |example|
    DatabaseCleaner.strategy = example.metadata[:js] ? :truncation : :transaction

    DatabaseCleaner.cleaning do
      example.run
    end
  end
end
```

Example provided by @lucashungaro fixes the issue, however I'm not sure why this happens. Any ideas?

lucashungaro:
@airatshigapov It's probably an issue with around filters not working properly, so the transactions are never closed.

ka8725:
[...]

PR #360, opened 2015-05-15
Add documentation for RSpec with Capybara

tgaff:
I think there are probably a large percentage of users that need DatabaseCleaner for Capybara specs.  I thought it would be nice to help them get going quickly with a little documentation.

Note: I could have used an around(:each) block as in the ordinary RSpec
example but there's currently an open issue that might affect many users:
https://github.com/DatabaseCleaner/database_cleaner/issues/273

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -219,6 +219,32 @@ RSpec.configure do |config|
 end
 ```
 
+### RSpec with Capybara Example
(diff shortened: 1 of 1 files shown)


etagwerker:
@tgaff Thank you!

PR #362, opened 2015-05-28
Add support for Firebird

rzane:
The adapter can be found here: https://github.com/rowland/activerecord-fb-adapter


etagwerker:
@rzane Thank you! Could you add a spec for it here: https://github.com/DatabaseCleaner/database_cleaner/tree/master/spec/database_cleaner/active_record/truncation? It should be a new file for this new database. 

Also, please rebase from `master`

PR #363, opened 2015-05-29
Adds support for additional Neo4j::Session.open options

brienw:
The current Neo4j support only passed through path and type parameters, but Neo4j::Session.open can also take in a third 'params' option for specifying things like connection profile or authentication. This change passes through the additional options to the Session.open call

### DIFF
--- a/lib/database_cleaner/neo4j/base.rb
+++ b/lib/database_cleaner/neo4j/base.rb
@@ -50,8 +50,12 @@ module DatabaseCleaner
         database[:path]
       end
 
+      def db_params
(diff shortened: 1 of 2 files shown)


etagwerker:
@brienw Thanks for this! Could you add a spec for this new behavior here: https://github.com/DatabaseCleaner/database_cleaner/blob/master/spec/database_cleaner/neo4j/base_spec.rb?

brienw:
@etagwerker Thanks, I must have been in a hurry when I opened this earlier. I've added spec coverage

etagwerker:
@brienw Awesome. Thank you! :+1:

PR #367, opened 2015-06-07
Added support for Mongoid 5.0 and Mongo Ruby Driver 2.0

jprincipe:


lib/database_cleaner/mongo2/truncation_mixin.rb
```diff
@@ -0,0 +1,37 @@
...
+        else
+          collections.each { |c| database[c].find.delete_many unless @tables_to_exclude.include?(c) }
+        end
+        true
+      end
+
+      private
+
+      def database
+        if not(@db.nil? or @db == :default)
```
etagwerker: @jprincipe Could you make this condition without the `not`? I think it'd make it easier to understand.

jprincipe: good call.  if/else logic should be clearer now

etagwerker:
@jprincipe Thanks!

Issue #345, opened 2015-03-11
column "table_rows" does not exist


coorasse:
Just use database_cleaner with Postgres and `DatabaseCleaner.strategy = :deletion`.
Version 1.4.1 is affected, version 1.4.0 I don't know. Version 1.3.0 is working fine.

ryanong:
This happens to us also. I think it may have something to do with the way schema_plus interacts with database_cleaner

coorasse:
No scherma plus for me.

casiodk:
+1

csdaraujo:
+1

mcclymont:
+1 With Postgres. Also have pg, postgres_ext, schema_plus gems. I get the issue on 1.4.1 and 1.4.0. We had upgraded to 1.4.1 as part of a Rails 3 -> 4 upgrade. I downgraded database_cleaner to 1.3.0 and this error went away.

shosti:
Also getting this. I'm not familiar with the inner workings of database_cleaner, but it looks like the problem is https://github.com/DatabaseCleaner/database_cleaner/blob/master/lib/database_cleaner/active_record/deletion.rb#L57 (introduced by https://github.com/DatabaseCleaner/database_cleaner/commit/81ca46565cf68a0e08fbfda7fb3542a119a2045d). `table_rows` doesn't exist in the Postgres information_schema AFAIK.

shosti:
Actually looks like this is to do with SchemaPlus (+ maybe other similar gems). From a rails app using PG:

```
[3] pry(main)> ActiveRecord::ConnectionAdapters::Mysql2Adapter
=> SchemaPlus::ForeignKeys::ActiveRecord::ConnectionAdapters::Mysql2Adapter
```

That's triggering the `defined?` condition. I'd be happy to look into this some more...

PR #373, opened 2015-07-06
Fix an issue when the mysql2 gem is included in a project that uses non-mysql databases.

scosman:
The old implementation assumed if the gem was present, that all databases were mysql. Connecting to a DB of any other type would cause exceptions. Instead, we check the connection adapter type in each instance of the cleaner, and only use the mysql2 specific code when using the mysql2 adapter.

### DIFF
--- a/lib/database_cleaner/active_record/deletion.rb
+++ b/lib/database_cleaner/active_record/deletion.rb
@@ -59,7 +59,7 @@ module DatabaseCleaner::ActiveRecord
     end
 
     def is_mysql2? connection
-      connection.class == 'ActiveRecord::ConnectionAdapters::Mysql2Adapter'
+      connection.class.to_s == 'ActiveRecord::ConnectionAdapters::Mysql2Adapter'
     end
 
     def clean
--- a/lib/database_cleaner/active_record/deletion.rb
+++ b/lib/database_cleaner/active_record/deletion.rb
@@ -43,35 +43,23 @@ module ActiveRecord
 end
 
 module DatabaseCleaner::ActiveRecord
-  module SelectiveTruncation
+  class Deletion < Truncation
     def tables_to_truncate(connection)
-      if information_schema_exists?(connection)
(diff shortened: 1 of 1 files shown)


lib/database_cleaner/active_record/deletion.rb
```diff
@@ -43,35 +43,23 @@ module ConnectionAdapters
...
-          false
-        end
-    end
-  end
-
-  class Deletion < Truncation
-    if defined?(ActiveRecord::ConnectionAdapters::Mysql2Adapter)
-      include SelectiveTruncation
+    def is_mysql2? connection
+      connection.class.to_s == 'ActiveRecord::ConnectionAdapters::Mysql2Adapter'
```
ahorek: jdbc mysql2 adapter (jruby) will return ActiveRecord::ConnectionAdapters::MysqlAdapter

scosman: That wouldn't have used SelectiveAdapter in the old implementation either. `defined?(ActiveRecord::ConnectionAdapters::Mysql2Adapter)` in the old code returns nil, even if jdbc adapter is loaded. 

Is the perf benefit really worth the conditional branching? An even cleaner fix than this patch would be to tear it all out.

ebeigarts:
This should also fix #345

lib/database_cleaner/active_record/deletion.rb
ebeigarts: I think it should check both, It doesn't make sense to leave it broken for jdbc:

```ruby
def is_mysql? connection
  adapter_name = connection.adapter_name
  adapter_name == 'Mysql2'.freeze || adapter_name == 'MySQL'.freeze
end
```

scosman: That looks fine with me, but is anyone actually willing to merge the PR?

etagwerker: @scosman Yes, I'd be willing to merge the PR if you add a spec with the scenario you're fixing.

PR #375, opened 2015-07-11
Less hacky way to obtain connection database name for deletion.

oggy:
In addition to violating encapsulation a little less, this allows
cleaning the correct shard when used with the popular octopus gem.

### DIFF
--- a/lib/database_cleaner/active_record/deletion.rb
+++ b/lib/database_cleaner/active_record/deletion.rb
@@ -53,7 +53,7 @@ module DatabaseCleaner::ActiveRecord
     end
 
     def tables_with_new_rows(connection)
-      @db_name ||= connection.instance_variable_get('@config')[:database]
+      @db_name ||= connection.pool.spec.config[:database]
       result = connection.exec_query("SELECT table_name FROM information_schema.tables WHERE table_schema = '#{@db_name}' AND table_rows > 0")
       result.map{ |row| row['table_name'] } - ['schema_migrations']
     end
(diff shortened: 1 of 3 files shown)


etagwerker:
@oggy I appreciate your contribution. Could you add coverage for the deletion strategy?

oggy:
@etagwerker Thanks, pushed!

As per the commit comment, however, I think these tests are pretty awful, and stub way too much to be very useful.

I started down the path of more integration-level tests -- i.e. hitting the actual database (using db/config.yml which I think is currently unused?) -- but in doing so I discovered that the gem versions in the gemspec actually aren't sufficient for the `ActiveRecord::Deletion` strategy. It calls `exec_query` which was only added in AR 3.1.0, and several gems in the lock file are tied to 3.0.0.

Trying to change the AR dependency version caused a cascade of bundler errors as other gems lock to AR 3.0. I think a better solution there would be to use different bundler groups or Gemfiles for the different major dependency combinations rather than using a single fixed set of development dependencies in the gemspec. That way you could also test against multiple versions of things.

All that said, doing that seems outside the scope of this tiny change. Let me know if you have other thoughts on the way forward here.

lostapathy:
@oggy is right on - would work well to take the dependencies out of the Gemfile.lock and use something like [Appraisal](https://github.com/thoughtbot/appraisal) to build a test matrix of various versions.

botandrose:
Thank you for the PR! However, the ActiveRecord adapter has been extracted into its own repo at https://github.com/DatabaseCleaner/database_cleaner-active_record. If you close this PR and reopen it over there, I'd be happy to take a look!

PR #377, opened 2015-07-14
 Explain & strengthen RSpec+Capybara config 

eliotsykes:
The configuration at https://github.com/eliotsykes/rspec-rails-examples/blob/master/spec/support/database_cleaner.rb is a tad more robust in 2 ways:

- It checks the Capybara driver isn't rack_test - this is the more reliable check. Using the `js: true` flag is only one of the ways in which a driver can be set. For example the `:driver` metadata option is an alternative way of setting the capybara driver.
- Setting `config.use_transactional_fixtures = false` in the database cleaner configuration is not always enough alone. If the config is still set in `rails_helper.rb` to `config.use_transactional_fixtures = true`, that can take precedence. The linked to configuration raises an error if this error is detected.

Alternatively I can submit a PR with the entire config in the README - though its quite a long config and felt a bit presumptive!

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -221,7 +221,30 @@ end
 
 ### RSpec with Capybara Example
 
-If you're using Capybara with RSpec and using an external browser (not using RackTest) you'll almost certainly need to use truncation rather than transactions for tests tagged `:js`.
+You'll typically discover a feature spec is incorrectly using transaction
+instead of truncation strategy when the data created in the spec is not
+visible in the app-under-test.
+
+A frequently occurring example of this is when, after creating a user in a
+spec, the spec mysteriously fails to login with the user. This happens because
+the user is created inside of an uncommitted transaction on one database
+connection, while the login attempt is made using a separate database
+connection. This separate database connection cannot access the
+uncommitted user data created over the first database connection due to
+transaction isolation.
+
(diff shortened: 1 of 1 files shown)


etagwerker:
@eliotsykes Thank you for your contribution! 

I'd prefer it if you just change the existing sample. Your sample is much more complete. 

I'd like to see https://github.com/eliotsykes/rspec-rails-examples/blob/master/spec/support/database_cleaner.rb#L18-L44 and https://github.com/eliotsykes/rspec-rails-examples/blob/master/spec/support/database_cleaner.rb#L18-L44 in the official sample. 

Could you update this PR to change the README accordingly?

eliotsykes:
Sounds good and thanks @etagwerker. Both links above point to lines 18-44, were 2 separate snippets of the sample intended?

etagwerker:
I meant the second link to be: https://github.com/eliotsykes/rspec-rails-examples/blob/master/spec/support/database_cleaner.rb#L54-L67 (copy/paste fail)

eliotsykes:
@etagwerker many thanks - updated PR and feedback welcome!

etagwerker:
@eliotsykes Great, thanks! :+1:

PR #382, opened 2015-08-06
remove non needed context args

HaiTo:
rspec `context`  is not needed args strings.

:+1:


etagwerker:
@HaiTo Thank you for your contribution. I'd prefer it if you add a descriptive string to the context. It'd make it easier to maintain it in the future. Could you do that?

HaiTo:
thank you responses.
bad, i cant easier to maintain :sweat: 

and so, it is that this is nothing and closeing. thank you :)

Issue #378, opened 2015-07-20
Mongoid removed the remove_all method?


gotjosh:
I tried setting up database_cleaner on a project using mongoid `5.0.0beta`, it seems they removed the `remove_all` method so it's no longer working.

Here's the error I'm getting:

```
/Users/gotjosh/.rbenv/versions/2.2.2/gemsets/aplus/gems/database_cleaner-1.4.1/lib/database_cleaner/moped/truncation_base.rb:14:in `block in clean': undefined method `remove_all' for #<Mongo::Collection::View:0x007fd6600c6c08> (NoMethodError)
	from /Users/gotjosh/.rbenv/versions/2.2.2/gemsets/aplus/gems/database_cleaner-1.4.1/lib/database_cleaner/moped/truncation_base.rb:14:in `each'
	from /Users/gotjosh/.rbenv/versions/2.2.2/gemsets/aplus/gems/database_cleaner-1.4.1/lib/database_cleaner/moped/truncation_base.rb:14:in `clean'
	from /Users/gotjosh/.rbenv/versions/2.2.2/gemsets/aplus/gems/database_cleaner-1.4.1/lib/database_cleaner/base.rb:40:in `clean_with'
	from /Users/gotjosh/rails/aplus/spec/rails_helper.rb:15:in `block (2 levels) in <top (required)>'
	from /Users/gotjosh/.rbenv/versions/2.2.2/gemsets/aplus/gems/rspec-core-3.2.2/lib/rspec/core/example.rb:333:in `instance_exec'
	from /Users/gotjosh/.rbenv/versions/2.2.2/gemsets/aplus/gems/rspec-core-3.2.2/lib/rspec/core/example.rb:333:in `instance_exec'
	from /Users/gotjosh/.rbenv/versions/2.2.2/gemsets/aplus/gems/rspec-core-3.2.2/lib/rspec/core/hooks.rb:357:in `run'
	from /Users/gotjosh/.rbenv/versions/2.2.2/gemsets/aplus/gems/rspec-core-3.2.2/lib/rspec/core/configuration.rb:1559:in `block in run_hooks_with'
	from /Users/gotjosh/.rbenv/versions/2.2.2/gemsets/aplus/gems/rspec-core-3.2.2/lib/rspec/core/configuration.rb:1559:in `each'
[...]

PR #385, opened 2015-08-16
Added mongo2/* files to gemspec so that Mongoid >= 5 would work.

geerzo:
This should let the changes from dd14ae4b762cf5f9a9edef2966ef1a24c754f8c3 work, addressing #378


etagwerker:
@geerzo Thanks for your contribution! Would you mind making it so that everything in `lib` gets added programmatically?

etagwerker:
@geerzo Thanks for your contribution! 

After merging this manually, I've improved the gemspec with this commit: https://github.com/DatabaseCleaner/database_cleaner/commit/dabfe2fcfb3678b92d48ce333739bda8e995ef68

geerzo:
@etagwerker yes, that solution is much better. I went with the "simple" solution assuming it was done that way for a reason.

PR #405, opened 2015-10-24
Add comparable support to strategies

st0012:
Add comparable as [This todo](https://github.com/DatabaseCleaner/database_cleaner/blob/master/lib%2Fdatabase_cleaner%2Fbase.rb#L99) mentioned.

### DIFF
--- a/lib/database_cleaner/base.rb
+++ b/lib/database_cleaner/base.rb
@@ -4,7 +4,7 @@ module DatabaseCleaner
     include Comparable
 
     def <=>(other)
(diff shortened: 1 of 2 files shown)


schmierkov:
@st0012 could you also provide test coverage for this change? Thanks in advance!

st0012:
@schmierkov ok, I have added some tests, and the CI fails because of bundle error.

etagwerker:
@st0012 @schmierkov Great, thanks! :+1:

PR #408, opened 2015-12-01
Use default_client following mongoid 5.0

stjhimy:
http://www.rubydoc.info/github/mongoid/mongoid/Mongoid%3Adefault_session


etagwerker:
@stjhimy Thanks! :+1:

PR #413, opened 2015-12-22
Redis deprecations

kosdiamantis:
* Replace `Redis.connect`[deprecated] with `Redis.new`
* Add ability to pass directly the redis connection instead of redis url.


etagwerker:
@kosdiamantis Thank you! :+1:

PR #416, opened 2015-12-24
update travis bundler before install

schmierkov:
Travis builds are currently failing on the new infrastructure.

```
NoMethodError: undefined method `spec' for nil:NilClass
```
and
```
An error occurred while installing
database_cleaner (1.5.1), and Bundler
cannot continue.
Make sure that `gem install
database_cleaner -v '1.5.1'` succeeds
before bundling.
```

As advised in https://github.com/travis-ci/travis-ci/issues/3531, it seems to be necessary to install a newer bundler version before running `bundle install`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,8 @@ rvm:
 script: "bundle exec rake"
 gemfile:
   - Gemfile
+before_install:
+  - gem update bundler
 before_script:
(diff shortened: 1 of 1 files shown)


etagwerker:
@schmierkov Great, thanks! :+1:

PR #417, opened 2015-12-29
Don't try to use mysql-specific information_schema

shosti:
Fixes https://github.com/DatabaseCleaner/database_cleaner/issues/345

I tried to add specs, but it looks like there currently aren't any real specs for the deletion strategy (correct me if I'm wrong), and ActiveRecord is locked at 3.0.0 which doesn't support `exec_query`. Updating everything seemed like a pretty major undertaking--I'd be happy to write some real specs once that's done.

### DIFF
--- a/lib/database_cleaner/active_record/deletion.rb
+++ b/lib/database_cleaner/active_record/deletion.rb
@@ -59,6 +59,7 @@ module DatabaseCleaner::ActiveRecord
     end
 
     def information_schema_exists? connection
+      return false unless connection.is_a? ActiveRecord::ConnectionAdapters::Mysql2Adapter
       @information_schema_exists ||=
         begin
           connection.execute("SELECT 1 FROM information_schema.tables")


etagwerker:
@shosti I appreciate your contribution, but could you add coverage for this scenario?

shosti:
Trying to add specs, but running into some snags:

1. The deletion strategy is currently pretty much untested AFAICT.
2. The dev Gemfile is locked at ActiveRecord 3.0.0.
3. AR 3 doesn't have `exec_query` which is currently used by the deletion strategy.
4. I thought about changing `exec_query` to `execute`, but the API changed between AR 3 and 4.
5. I tried updating AR to 4.2.5, but a large number of existing specs fail in unexpected ways with AR 4 (see https://gist.github.com/e98dc0133843d91fe79d for an example). 

In other words, it looks like there are some deeper issues with the test suite at the moment. I'd be happy to help patch it up with some guidance, but probably not if it involves rewriting significant portions of the test suite.

lastobelus:
+1 This patch is currently needed to use 1.4+ with schema_plus

etagwerker:
@shosti Ok, thanks for your contribution! You're right. I tried to do this myself and I think it'd be best to wait until 2.0 is released. 2.0 should allow us to have better dependency management.

PR #431, opened 2016-03-31
In specs, drop Postgres database with AR connection

skalee:
1. Works faster than issuing shell commands.
2. Attempts to drop the database before creating it.  This fixes some ugly test output which warns that given database already exists.


etagwerker:
@skalee Great, thanks! 👍

PR #432, opened 2016-03-31
Upgrade RSpec to 2.14

skalee:
DBCleaner relies on some horribly old RSpec-Rails version. It's not even the most recent 2.x one. Let's fix it.


etagwerker:
@skalee Great, thanks! 🍻

PR #434, opened 2016-04-02
Escape asterisk in the README

skalee:
Markdown used to interpret it as one-element list.


etagwerker:
@skalee Great, thanks!

PR #437, opened 2016-04-18
Detect mongoid version to use correct method.

P9GIT:
"default_session" is replaced with "default_client" in mongoid version >= 5.0.0

This would make database_cleaner run with older versions of Mongoid.


etagwerker:
@P9GIT Great, thanks! 👍

PR #452, opened 2016-07-08
Fix redis db option

soylent:


cgunther:
I hit the same issue with v1.5.3 and this PR fixed it for me.

The error I got when running my tests was:
```
Redis::CommandError:
  ERR unknown command 'to_str'
```

If you're assigning an instance of `Redis`, it's class will not be an instance of `Redis`, as the code originally checked for. This caused it to create a new `Redis` instance, passing the assigned instance as if it were a URL string.

etagwerker:
@soylent Thanks for your contribution! 👍

PR #457, opened 2016-08-30
Adding support for neo4j-core 7.0.x (neo4j 8)

ProGM:
Hi!
Soon there will be a release of `neo4j-core` 7.0.x (currently in alpha), that changes entirely the connection/session API, and comes with the new `BOLT` protocol support.

There will be many breaking changes on how sessions are created, that influences how `database_cleaner` would work with it.

For example:
There's no more a `current_session`, because you can have multiple sessions using different adaptors. As a consequence, running queries and transactions requires to have an instance of a `::Neo4j::Core::CypherSession` object

The "current_session" concept was moved to [`neo4j`](https://github.com/neo4jrb/neo4j) gem, the ActiveRecord-like ORM for neo4j. If you're using it, the most simple way to migrate settings would be to change this:

``` ruby
    DatabaseCleaner[:neo4j, connection: {
      type: Rails.application.config.neo4j.session_type,
      path: Rails.application.config.neo4j.session_path
    }]
```

into this:

``` ruby
    DatabaseCleaner[:neo4j, connection: { current_session: Neo4j::ActiveBase.current_session}]
```

If you're using `neo4j-core` without it, instead, you can pass to `DatabaseCleaner` your opened session to `current_session` or also the connection settings.

I made a few changes to `DatabaseCleaner`, so it should support both `neo4j-core < 7.0` and `>= 7.0`

P.S. I'm still working on this 😜

### DIFF
--- a/lib/database_cleaner/neo4j/base.rb
+++ b/lib/database_cleaner/neo4j/base.rb
@@ -1,4 +1,6 @@
 require 'database_cleaner/generic/base'
+require 'neo4j-core'
+
 module DatabaseCleaner
   module Neo4j
     def self.available_strategies
@@ -13,11 +15,11 @@ module DatabaseCleaner
       end
 
       def db
-        @db ||= nil
+        @db
       end
 
       def start
-        if db_type == :embedded_db and not session.running?
+        if db_type == :embedded_db && !session.running?
           session.start
         else
           session
@@ -26,6 +28,14 @@ module DatabaseCleaner
 
       def database
         db && default_db.merge(db) || default_db
+      end
+
+      def query(*args)
+        if legacy_neo4j?
+          session._query(*args)
+        else
+          session.query(*args)
+        end
       end
 
       private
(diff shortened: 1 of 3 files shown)


etagwerker:
@ProGM Thanks for this! Have you been able to make any more progress? 

Please let me know when it's ready for review.

ProGM:
@etagwerker Currently it's working, but I'm waiting for an official release fro neo4j, so I can write some specs on it ;)

tseleb:
We would love to have this as DatabaseCleaner is deleting our database `SchemaMigrations`!

cheerfulstoic:
Was working with @tseleb on this and after thinking about it we think that the solution might actually be creating a Neo4j alternative to the `schema.rb`.  In `ActiveRecord` (if I remember correctly) the `db/schema.rb` is used to set up the schema of the database at the start of the test suite.  The `schema.rb` is also generated automagically after the migrations run based on what's in the database.

What's more, we need to have a way to be able to run tests with the schema migrations run but not the data migrations.  The setup that is done in the tests (be it fixtures or factories) should represent the current understanding of what the data should be after any migrations have been run.  Because of that I don't think we have any alternative but to create a `schema.rb` alternative.

Thoughts @ProGM ?

cheerfulstoic:
(also, that said, I don't think that `check_for_pending_migrations` should be run in test mode)

olleolleolle:
(Yay, 7.0 of the upstream gem was released.)

dominicsayers:
Please merge! :shipit:

cheerfulstoic:
@dominicsayers There isn't much benefit (currently) to using the `database_cleaner` gem with Neo4j because it doesn't support truncation or nested transaction.  Check out this section in the docs for more info:

http://neo4jrb.readthedocs.io/en/8.0.x/Miscellany.html#cleaning-your-database-for-testing

botandrose:

olleolleolle:
Right, and that repo was just archived. https://github.com/DatabaseCleaner/database_cleaner-neo4j

botandrose:
Yeah, more context here: https://github.com/DatabaseCleaner/database_cleaner/issues/647

PR #465, opened 2017-01-01
Fix truncation method for old versions of mongodb

zoso10:
The mongo gem removed the `Mongo::VERSION` constant for versions [1.8.x](https://github.com/mongodb/mongo-ruby-driver/blob/1.8.0/VERSION) to 1.12.x till it returned in version 2 and above. For super old versions of mongo, namely [1.7.x](https://github.com/mongodb/mongo-ruby-driver/blob/1.7.1/lib/mongo/version.rb) and below the constant still existed; so we need to check definition of the version constant and check the version number to correctly select the truncation method name. Also, added a spec to exercise the bug that previously existed.

### DIFF
--- a/lib/database_cleaner/mongo/truncation_mixin.rb
+++ b/lib/database_cleaner/mongo/truncation_mixin.rb
@@ -18,8 +18,8 @@ module DatabaseCleaner
       end
 
       def truncate_method_name
-        # This constant only exists in the 2.x series.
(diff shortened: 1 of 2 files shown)


botandrose:
Thank you for the PR! However, the Mongo adapter has been extracted into its own repo at https://github.com/DatabaseCleaner/database_cleaner-mongo. If you close this PR and reopen it over there, I'd be happy to take a look!

PR #472, opened 2017-01-21
Random ids

randoum:
I sometimes (often ?) mistakenly use a wrong record at the wrong place, and the specs could wrongly be all green only because in the running test all records `id` have the same value : 1. 

This PR is to have DatabaseCleaner to set auto-increment/sequence to a random value when calling `DatabaseCleaner.clean_with`

- Works with MySql and PostgreSql
- Works with truncation only
- Disabled by default (though enabling it by default could be a good idea, but it's time consuming)
- Can be set for all tables or only for specific tables
- Can specify the primary key for each table (useful only for PostgreSal)
- Cannot be used alongside with `:pre_count`

Activated by calling `DatabaseCleaner.clean_with :truncation, random_ids: true`

`random_ids: %w[table1 table2]` would set random auto-increment/sequence to table1 and table2 _only_, all other table would be truncated as usual

`random_ids: {table1: 'id', table2: 'key'}` would set random auto-increment/sequence to table1 and table2 _only_, AND specifying that table2's primary key's name is 'key'. All other table would be truncated as usual

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -211,6 +211,24 @@ The following option is available for ActiveRecord's `:truncation` and `:deletio
 
 * `:cache_tables` - When set to `true` the list of tables to truncate or delete from will only be read from the DB once, otherwise it will be read before each cleanup run. Set this to `false` if (1) you create and drop tables in your tests, or (2) you change Postgres schemas (`ActiveRecord::Base.connection.schema_search_path`) in your tests (for example, in a multitenancy setup with each tenant in a different Postgres schema). Defaults to `true`.
 
+### Setting random ids when cleaning database
+
+The following is available for ActiveRecord's :truncation strategy _only_ for MySQL and Postgres.
+
+It may be useful to initialize your database with a random value for auto-increment (in case you are using MySQL) or for sequences (in case you are using Postgres).
+Simply call `clean_with` with the `:random_ids` option:
+
+```
+DatabaseCleaner.clean_with :truncation, random_ids: true
+```
+
+`:random_ids` accept the following values:
+
+* `true`: sets a random value for every table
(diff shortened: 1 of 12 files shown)


randoum:
@etagwerker could you spare some time and have a look ?

etagwerker:
@randoum First of all thank you for taking the time to submit a contribution to this gem. 

Regarding the problem that you mention: 

> I sometimes (often ?) mistakenly use a wrong record at the wrong place, and the specs could wrongly be all green only because in the running test all records id have the same value : 1.

I don't see it as a problem that should be solved by `database_cleaner` - It might be a common problem for some, but I've never encountered it in my projects. 

The quick solution would be to not use ids to load required objects for your tests. If you need to load a specific country object, you could do something like this: 

```ruby
country = Country.find_by(name: 'Argentina')
```

not

```ruby
country = Country.find 123
```

Another alternative would be to have fixtures that set the id to always be the same one. 

Another alternative would be to use some sort of factory object to create such records.

I think it would be best if you wrote a new gem that depends on `database_cleaner`; something like `database_cleaner-random_ids`. 

So, if someone does want this problem, they can just include your gem and get this behavior.

randoum:
Well ... I was just trying to make a funny introduction. Don't be so serious :)

Thanks for your alternatives suggestions. You are saying to call `find_by` instead of `find`? Where did you get such an idea? Man, I have to rewrite all my routes, all my controller, all my select tags...
Your suggestion are no good buddy, I hope you don't get thins kind of inspiration with your clients. 

I'm using a simple rspec helper since a long time, just wanted to share it with the community. You don't want it, I won't insist.

PR #481, opened 2017-03-16
Fix node deletion for Neo4j

qbart:
Removes all nodes/relations except migrations.

### DIFF
--- a/lib/database_cleaner/neo4j/deletion.rb
+++ b/lib/database_cleaner/neo4j/deletion.rb
@@ -8,7 +8,7 @@ module DatabaseCleaner
 
       def clean
         ::Neo4j::Transaction.run do
-          session._query('MATCH (n) OPTIONAL MATCH (n)-[r]-() DELETE n,r')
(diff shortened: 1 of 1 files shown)


cheerfulstoic:
I was thinking about this, though if `database_cleaner` is used for just the `neo4j-core` gem then the extra `WHERE` isn't necessary.  I've actually *just* made a new section in the docs about using `database_cleaner`:

http://neo4jrb.readthedocs.io/en/8.0.x/Miscellany.html#cleaning-your-database-for-testing

I was actually coming here because I was thinking of making a PR to remove support in `database_cleaner` for Neo4j for transaction and truncation strategies because I think it just causes confusion otherwise (see the docs link as well as https://github.com/neo4jrb/neo4j/issues/1368)

The other thing is that the next thing that I would like to do is make tests not use migrations but rather use something like `ActiveRecord`'s `schema.rb`.  This would remove the need for the extra `WHERE` here.

Thoughts?

botandrose:

PR #484, opened 2017-04-27
innodb support (mysql 5.6+)

### LINKED ISSUE #455: tables_with_new_rows query isn't safe on innodb
I found out, that randomly my tables aren't cleaned when using deletion strategy on mysql. 
[
connection.exec_query("SELECT table_name FROM information_schema.tables WHERE table_schema = '#{@db_name}' AND table_rows > 0")](https://github.com/DatabaseCleaner/database_cleaner/blob/master/lib/database_cleaner/active_record/deletion.rb#L57)

According to the documentation:
For InnoDB tables, the row count is only a rough estimate used in SQL optimization.
http://dev.mysql.com/doc/refman/5.7/en/tables-table.html

Env:
MRI 2.1.5 & Rails 4.2.6
Mysql 5.7.14 winx64

ahorek:
fixes #455

1/ the original algorithm works only on mysql < 5.6 or if myisam engine is enabled. On innodb, deletion strategy isn't reliable at all. We have many test failures because of this bug.
2/ instead of using information_schema COUNT(*) entries from all tables. This is suprisingly much faster (about 30 times) and reliable.
3/ also accessing information_schema is VERY slow, if caching is enabled, we don't have to access information_schema over and over again. It saves some additional time between tests.

### DIFF
--- a/lib/database_cleaner/active_record/deletion.rb
+++ b/lib/database_cleaner/active_record/deletion.rb
@@ -54,8 +54,27 @@ module DatabaseCleaner::ActiveRecord
 
     def tables_with_new_rows(connection)
       @db_name ||= connection.instance_variable_get('@config')[:database]
-      result = connection.exec_query("SELECT table_name FROM information_schema.tables WHERE table_schema = '#{@db_name}' AND table_rows > 0")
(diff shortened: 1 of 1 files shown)


etagwerker:
@ahorek Thank you! 👍

PR #485, opened 2017-05-08
Fix missing method in deletion

kawamoto:
`migration_table_name` is not defined in this module


etagwerker:
@kawamoto Great, thanks! 👍

PR #490, opened 2017-06-01
Use the right ActiveRecord::Base module.

mauro-oto:
Hey @etagwerker, 

This PR fixes a problem with database_cleaner and Rails 5. The module name syntax for `ActiveRecord::Base` in the exclusion condition was incorrect. 

Please check it out, thanks!

PR #491, opened 2017-06-01
Fix ActiveRecord::Base.internal_metadata_table_name lookup

stevenharman:
Without this, `ActiveRecord::Base.internal_metadata_table_name` resolves to the `DatabaseCleaner::ActiveRecord::Base`, which doesn't have that method. This means trying to run with a truncation strategy on Rails >= 5 results in the following error:

```
NoMethodError:
    undefined method `internal_metadata_table_name' for DatabaseCleaner::ActiveRecord::Base:Module
```

Phew!


etagwerker:
@stevenharman Thank you! I decided to go with #490. 👍

PR #496, opened 2017-07-26
Quote table name in ActiveRecord::SelectiveTruncation#table_stats_query

odedniv:
Kind of patchy, maybe someone can think of a better way.

My tables need quoting, and there's a change in 1.6 that didn't quote. It's a query generated with a string inside the query, so quoting is not straight forward (`table_name` is not a Ruby variable).

### DIFF
--- a/lib/database_cleaner/active_record/deletion.rb
+++ b/lib/database_cleaner/active_record/deletion.rb
@@ -66,8 +66,9 @@ module DatabaseCleaner::ActiveRecord
       if @cache_tables && !@table_stats_query.nil?
         return @table_stats_query
       else
+        quote_start, quote_end = connection.quote_table_name('T').split('T')
         @table_stats_query = connection.select_values(<<-SQL).join(' UNION ')
-               SELECT CONCAT('SELECT \"', table_name, '\" AS table_name, COUNT(*) AS exact_row_count FROM ', table_name)
+               SELECT CONCAT('SELECT \"', table_name, '\" AS table_name, COUNT(*) AS exact_row_count FROM #{quote_start}', table_name, '#{quote_end}')
                FROM
                INFORMATION_SCHEMA.TABLES
                WHERE


lib/database_cleaner/active_record/deletion.rb
```diff
@@ -66,8 +66,9 @@ def table_stats_query(connection, db_name)
       if @cache_tables && !@table_stats_query.nil?
         return @table_stats_query
       else
+        quote_start, quote_end = connection.quote_table_name('T').split('T')
         @table_stats_query = connection.select_values(<<-SQL).join(' UNION ')
-               SELECT CONCAT('SELECT \"', table_name, '\" AS table_name, COUNT(*) AS exact_row_count FROM ', table_name)
+               SELECT CONCAT('SELECT \"', table_name, '\" AS table_name, COUNT(*) AS exact_row_count FROM #{quote_start}', table_name, '#{quote_end}')
```
etagwerker: @odedniv This implementation looks more complicated than it needs to be. Could you see how this was implemented here: https://github.com/DatabaseCleaner/database_cleaner/blob/master/lib/database_cleaner/active_record/deletion.rb#L17

Is there a good reason you didn't just use `quote_table_name`?

odedniv: Yes, take a closer look. `table_name` is not a Ruby variable, inside a Ruby code, that I can use a Ruby function to manipulate. It's an SQL variable used in an SQL function (`CONCAT`) that generates an SQL string. That SQL string needs to have quotes inside it.

I tried `connection.quote_table_name('table_name')` hoping it would affect the SQL string, but it didn't.

buehmann: Generating the SQL strings in the database is really ugly. :smile: If you pull that bit into Ruby land you don't have trouble quoting properly, e.g. like this: #501

odedniv: Agreed, if you're OK with generating the SQL in Ruby (not sure about the reasoning behind CONCATing in SQL), I prefer #501 and this PR can be closed.

odedniv:
Closed by #501

PR #498, opened 2017-08-05
Check correct MySQL database during pre_count

### LINKED ISSUE #293: MySQL truncation with pre_count: true truncates unused tables in certain situations
I use a single MySQL installation for the development and testing of a Rails project, which contains both the development and test databases for the project. Naturally, both of these databases have the same tables. Most of my development database tables are populated with data.

I have quite a few tables, so I decided to set database cleaner to use the `truncation` strategy with 'conditional' truncation using the `pre_count: true` option. So, my RSpec `spec_helper` contains the following:

```
  config.before(:suite) do
    DatabaseCleaner.clean_with(:truncation)
    DatabaseCleaner.strategy = [:truncation, pre_count: true]
  end

  config.after(:each) do
    DatabaseCleaner.clean
  end
```

When I ran a spec file that only touches a few tables as a test for this new configuration, I noticed that a lot of tables were still being truncated despite not having been used in my tests.

After some investigation, I believe this behaviour is due to the MySQL query used to determine whether a table needs to be truncated or not based on its autoincrement value. The current query (as per https://github.com/DatabaseCleaner/database_cleaner/blob/master/lib/database_cleaner/active_record/truncation.rb#L73) looks up a given table name in the `information_schema.tables` table, which contains table information for all databases in the entire installation. It does not do any filtering based on the particular database the tests are being run in.

For my setup, the query results in two rows being returned - one for the table in the development database (which usually has an autoincrement value larger than one because most tables are populated with data), and one for the table in the test database (which will have an autoincrement value of 1 if it hasn't been used). I'm guessing that the autoincrement value for the table in the development database is the one that is actually used within database cleaner, possibly because that database was created before the test database and/or the development database name is alphabetically prior to the test database name. As a result of this, all tables in the test database for which data exists in the table of the same name in the development database are being truncated, regardless of whether or not they actually had any data created in them.

The fix I'm suggesting is to modify the query mentioned above to select only rows from the `information_schema.tables` table for the database in which the tests are being run. This should be able to be easily done via the `table_schema` column in that table and the value returned by the `current_database` method within database cleaner. As an example, a similar approach is used within database cleaner here:

https://github.com/DatabaseCleaner/database_cleaner/blob/master/lib/database_cleaner/active_record/truncation.rb#L23

So, the new query would be:

```
          SELECT Auto_increment
          FROM information_schema.tables
          WHERE table_schema = '#{current_database}'
          AND table_name='#{table}'
```

I monkey patched the appropriate method in `DatabaseCleaner::ActiveRecord::Truncation` to test this locally, and it worked. However, I'm afraid I have no idea how to write specs for this, so I haven't created a PR. If you agree that this is an issue and would like me to take any particular actions, please do let me know!

buehmann:
This fixes #293.

Truncation with `pre_count` takes wrong decisions for MySQL databases: It either truncates too many tables (bad for performance) or too few.

When inspecting the `auto_increment` value, it picks a random table out of many with the same name in different schemas/databases:
```
mysql> select table_schema, table_name, auto_increment from information_schema.tables where table_name = 'accounts';
+---------------------+------------+----------------+
| table_schema        | table_name | auto_increment |
+---------------------+------------+----------------+
| banking_development | accounts   |             15 |
| banking_profile     | accounts   |              1 |
| banking_test        | accounts   |              1 |
| performance_schema  | accounts   |           NULL |
+---------------------+------------+----------------+
4 rows in set (0.00 sec)
```
We want it to look for the right table (in the current database) instead.

### DIFF
--- a/lib/database_cleaner/active_record/truncation.rb
+++ b/lib/database_cleaner/active_record/truncation.rb
@@ -63,22 +63,21 @@ module DatabaseCleaner
         select_value("SELECT EXISTS (SELECT 1 FROM #{quote_table_name(table)} LIMIT 1)").to_i
       end
 
-      # Returns a boolean indicating if the given table has an auto-inc number higher than 0.
-      # Note, this is different than an empty table since an table may populated, the index increased,
-      # but then the table is cleaned.  In other words, this function tells us if the given table
-      # was ever inserted into.
-      def has_been_used?(table)
-        if has_rows?(table)
-          true
-        else
-          # Patch for MysqlAdapter with ActiveRecord 3.2.7 later
-          # select_value("SELECT 1") #=> "1"
-          select_value(<<-SQL).to_i > 1 # returns nil if not present
-          SELECT Auto_increment
+      def auto_increment_value(table)
+        select_value(<<-SQL).to_i
+          SELECT auto_increment
           FROM information_schema.tables
-          WHERE table_name='#{table}';
-          SQL
-        end
+          WHERE table_name = '#{table}'
+          AND table_schema = database()
+        SQL
+      end
+
+      # This method tells us if the given table has been inserted into since its
+      # last truncation. Note that the table might have been populated, which
+      # increased the auto-increment counter, but then cleaned again such that
+      # it appears empty now.
+      def has_been_used?(table)
+        has_rows?(table) || auto_increment_value(table) > 1
       end
 
       def has_rows?(table)


lib/database_cleaner/active_record/truncation.rb
```diff
@@ -74,9 +74,10 @@ def has_been_used?(table)
           # Patch for MysqlAdapter with ActiveRecord 3.2.7 later
           # select_value("SELECT 1") #=> "1"
           select_value(<<-SQL).to_i > 1 # returns nil if not present
-          SELECT Auto_increment
+          SELECT auto_increment
           FROM information_schema.tables
-          WHERE table_name='#{table}';
+          WHERE table_name = '#{table}'
+          AND table_schema = database()
```
etagwerker: @buehmann Good idea. Would you mind refactoring the SQL query into its own method?

buehmann: Sure, no problem.

buehmann:
@etagwerker I updated this PR a while ago following your comment. Do you have time to revisit it?

dvandersluis:
Will this ever be merged in? I ran into this today and I'd rather not monkey patch :smile:

etagwerker:
@buehmann Thanks for your contribution! 

@dvandersluis You can test with `master` until I release a new version of the gem.

dvandersluis:
Thank you @etagwerker!

PR #507, opened 2017-11-02
Fix Mongo truncation by improving Mongo version detection

laserlemon:
The `Mongo::VERSION` constant does not just exist since version 2.0 [as advertised](https://github.com/mongodb/mongo-ruby-driver/blob/v2.5.0.beta/lib/mongo/version.rb#L19-L20). It actually existed in all patch versions of 1.6 and [1.7](https://github.com/mongodb/mongo-ruby-driver/blob/1.7.1/lib/mongo/version.rb#L2) as well.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,1 +1,3 @@
+source "https://rubygems.org"
+
 gemspec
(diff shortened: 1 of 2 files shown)


laserlemon:
The failures are unrelated and seem to correspond exactly with [failures currently on `master`](https://travis-ci.org/DatabaseCleaner/database_cleaner/builds/296263717).

botandrose:
Thank you for the PR! However, the Mongo adapter has been extracted into its own repo at https://github.com/DatabaseCleaner/database_cleaner-mongo. If you close this PR and reopen it over there, I'd be happy to take a look!

Issue #506, opened 2017-10-31
Remove symlimk from package


etagwerker:
For more info: https://github.com/DatabaseCleaner/database_cleaner/commit/8c14a276ddb8115bb391d59cdf6b706577d29678#commitcomment-25314056

akatriel:
Still getting this issue 
```
Errno::EACCES: Permission denied @ rb_file_s_symlink - (examples/../Gemfile,
C:/Ruby23-x64/lib/ruby/gems/2.3.0/gems/database_cleaner-1.6.2/examples/Gemfile)
An error occurred while installing database_cleaner (1.6.2), and
Bundler cannot continue.
Make sure that `gem install database_cleaner -v '1.6.2'` succeeds before
bundling.
```

cleicar:
@etagwerker, I think that I fixed it on this commit:  https://github.com/DatabaseCleaner/database_cleaner/commit/a615552196d5707752d5e15752fe19e9b4c0badd

Did you release a new version? Can we close this issue?

etagwerker:
@akatriel @cleicar Could you test with `master` in Windows and confirm that you're no longer getting the `Errno::EACCES` error?

Honest-Objections:
Not sure how to run with ```master```, however the standard ```gem 'database_cleaner'``` is till installing ```1.6.2``` and failing

cleicar:
@etagwerker I came to the same conclusion. Changes are on `master` but the version isn't released on RubyGems. The version is still on `1.6.2`

rameshjesswani:
I am also facing same issue, I am on ```master```, but still it is not working for me. Can anyone help please?

```
ERROR:  While executing gem ... (Errno::EACCES)
    Permission denied @ rb_file_s_symlink - (examples/../Gemfile, C:/Ruby24/lib/
ruby/gems/2.4.0/gems/database_cleaner-1.6.2/examples/Gemfile)
```

Honest-Objections:
Simply use version ```1.6.1``` until this issue is resolved

gazayas:
Running windows 7, putting `gem 'database_cleaner', '~> 1.6.1' ` still throws an error saying I saying permission denied, make sure `gem install database_cleaner -v '1.6.2'` installs first.

I changed to the version to `1.5.0` in my gemfile and it finally worked (it installed `1.5.3`).

starrychloe:
[...]

PR #508, opened 2017-11-03
Remove unnecessary folders from gemspec

cleicar:
@etagwerker this PR is fixing the issue #506


etagwerker:
@cleicar Thank you! 👍

PR #448, opened 2016-06-03
Update to json 1.8.3 to fix issue for bundle install.

junaruga:
Hi,
## Summary

I failed to do "bundle install" in my environment.
So, I update dependency gem package json from 1.8.1 to latest version: 1.8.3 in Gemfile.lock.

I just updated below 3 lines in Gemfile.lock.
However other 2 line were updated, when I succeeded to "bundle install".
I used bundler latest version: 1.12.5 for that.

```
-    json (1.8.1)
-    json (1.8.1-java)
-    json_pure (1.8.1)
+    json (1.8.3)
+    json (1.8.3-java)
+    json_pure (1.8.3)
```
## Detail

My environment is Fedora 23.

```
$ gcc --version
gcc (GCC) 5.3.1 20160406 (Red Hat 5.3.1-6)
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

```
$ ruby -v
ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-linux]
```

```
$ bundle install --path vendor/bundle
...
Installing json 1.8.1 with native extensions

Gem::Ext::BuildError: ERROR: Failed to build gem native extension.

    current directory: /home/jaruga/git/database_cleaner/vendor/bundle/ruby/2.3.0/gems/json-1.8.1/ext/json/ext/generator
/usr/local/ruby-2.3.1/bin/ruby -r ./siteconf20160603-11142-q93i3x.rb extconf.rb
creating Makefile

current directory: /home/jaruga/git/database_cleaner/vendor/bundle/ruby/2.3.0/gems/json-1.8.1/ext/json/ext/generator
make "DESTDIR=" clean

current directory: /home/jaruga/git/database_cleaner/vendor/bundle/ruby/2.3.0/gems/json-1.8.1/ext/json/ext/generator
make "DESTDIR="
compiling generator.c
In file included from generator.c:1:0:
../fbuffer/fbuffer.h: In function ‘fbuffer_to_s’:
../fbuffer/fbuffer.h:175:47: error: macro "rb_str_new" requires 2 arguments, but only 1 given
     VALUE result = rb_str_new(FBUFFER_PAIR(fb));
                                               ^
../fbuffer/fbuffer.h:175:20: warning: initialization makes integer from pointer without a cast [-Wint-conversion]
[...]

Issue #488, opened 2017-05-18
Remove Gemfile.lock


etagwerker:
Here is [why](http://stackoverflow.com/a/4151540/615138).

botandrose:
Resolved by d6f8bb133fd2546adc41958790f183be4d7f056a

PR #514, opened 2017-12-21
Fix travis build error and expand test matrix

lostapathy:
Attempting to fix the error leading to broken travis build and add ruby 2.3 to the test matrix.

This also updates the json dep so that more recent versions of ruby can build this gem - resolves #448

As is noted on #488, it's really not best practice to keep Gemfile.lock in the repo for a gem.  Looks like maintainers have changed - are we open to that now?  If so, I would work on PR for that and on expanding the test matrix to test this against various rails versions.  Thanks


lostapathy:
Well - this fixes the immediate error Travis was having, but there are still some (pre-existing, I believe) failing tests.  I'll try to loop around and look at those, but hopefully this PR is at least a little help.

etagwerker:
@lostapathy Thanks for your contribution! 

> As is noted on #488, it's really not best practice to keep Gemfile.lock in the repo for a gem. Looks like maintainers have changed - are we open to that now?

Yes, we're open to that. I started working on that but had to focus on something else. It'd be great if you can work on a PR for that.

lostapathy:
I'll take a stab at it.  @etagwerker how far back to you want to maintain support in the matrix?  I try not to include unsupported ruby or rails versions unless there's a really compelling reason, but looks like the current build goes back to ruby-1.9.3 and rails-3.0

etagwerker:
@lostapathy I think it's a good idea to move towards supported ruby/rails versions.

etagwerker:
@lostapathy I just took at stab at this and I now remember why it is so hard. `database_cleaner` has a lot of development dependencies. I believe it's way too many for any maintainer.

So, this task seems quite hard for `database_cleaner` < 2.0. Last year I started working on `database_cleaner` 2.0 but then got side tracked and couldn't finish it. If you're open to new contribution ideas, your time might be best spent on `database_cleaner` 2.0. 

The idea behind 2.0 is to have smaller gems with more manageable dependencies (e.g. database_cleaner-active_record)

In these smaller gems, we will be able to add `appraisals` to make it easier to test different dependencies (e.g. activerecord 4.2, 5.0, and 5.1)

lostapathy:
I removed the Gemfile lock and bundler takes ~45 minutes to resolve all the development dependencies.  

Yes - I agree 100%, the right way forward is to split this into several smaller gems, like a "core" + strategies for each database.

Issue #511, opened 2017-11-16
Getter method for strategy (to help with debugging)


rlue:
It appears that DatabaseCleaner does not currently support a getter method for determining which strategy it currently employs:

```ruby
>> DatabaseCleaner.strategy
NoMethodError: undefined method `strategy' for DatabaseCleaner:Module
Did you mean?  strategy=
from (pry):1:in `block (2 levels) in <top (required)>'
```

It would be nice to have this just so I could confirm, in my debugger, that I have DatabaseCleaner set up the way I had intended.

Is there another way to do this that I've missed?

etagwerker:
@rlue I think it's a good idea but it's not a trivial change because of this: 

https://github.com/DatabaseCleaner/database_cleaner/blob/f052d64d3be9351a98a99e932a59be528ccc561c/lib/database_cleaner/configuration.rb#L64-L67

If your application is using more than one connection, then that could return an Array (or a Hash?) 

Could you work on a PR for this?

rlue:
I am, but it's going to be a while before I have time to familiarize myself with the codebase enough to know what the implications of my changes will be.

I'm not sure I understand — the `::strategy=` method appears to unconditionally set the strategy on all connections. Is there another way that an individual connection's strategy might be changed? Otherwise, wouldn't we expect all connections to share the same strategy?

rlue:
@etagwerker I finally got around to taking a look at the code. Here's my understanding — please correct me if I'm mistaken:

* Each “connection” is an element of the `@connections` instance variable on the `DatabaseCleaner` module, and
* These connections are instances of the `DatabaseCleaner::Base` class.

What would you think of adding the `#strategy` getter method on the `DatabaseCleaner::Base` class? That way, `DatabaseCleaner[].strategy` would return the strategy for the default connection, and `DatabaseCleaner[:mongoid].strategy` would return the strategy for that one. Then, for some syntactic sugar, we could add a `DatabaseCleaner.strategy` method that aliases `DatabaseCleaner[].strategy`.

On the other hand, judging from the comments, it looks like you plan on doing a deeper redesign of the class hierarchy database_cleaner. Thoughts?

rlue:
Sorry, disregard my last comment. 1) I hadn't noticed that `DatabaseCleaner::Base#strategy` is already defined, and 2) I misunderstood how `DatabaseCleaner::[]` works. 

I'm submitting a PR according to your previous suggestion.

PR #518, opened 2018-01-06
Add DatabaseCleaner::strategy method

rlue:
This PR adds a `DatabaseCleaner::strategy` getter method as discussed in issue #511. 

Since there may be multiple connections, and different strategies for each connection, this getter method returns a hash of strategies.

The guard clause may be unnecessary, since the `::connections` method already attempts to autodetect an ORM and populate the `@connections` array with it, but I figured better safe than sorry.


rlue:
Whoops, I can't get the mysql gem to install on my machine. It's been more or less unmaintained for the past three years, and according to [this SO answer](https://stackoverflow.com/a/41521197/4865822), it's not compatible with Ruby > 2.3.

Any chance you'd be open to switching to [mysql2](https://github.com/brianmario/mysql2) to enable development in more modern versions of Ruby?

rlue:
Ayayay, never mind; I see that mysql2 is already a development dependency as well.

Are you able to work on this library in a currently-supported version of Ruby? I went ahead and installed Ruby 2.3.0, but now `bundle install` is borking on json v1.8.1:

```
An error occurred while installing json (1.8.1), and Bundler cannot continue.
Make sure that `gem install json -v '1.8.1'` succeeds before bundling.

In Gemfile:
  couch_potato was resolved to 1.3.0, which depends on
    couchrest was resolved to 1.2.0, which depends on
      rest-client was resolved to 1.6.8, which depends on
        rdoc was resolved to 4.1.2, which depends on
          json
```

and according to [this SO answer](https://stackoverflow.com/a/29578293/4865822), Ruby 2.2 broke json v1.8.1 (thus json >= 1.8.2 is required for Ruby >= 2.2).

I'm open to making any necessary modifications to the library to bring it up to date, but I'm reluctant to undertake the work without your blessing.

botandrose:
@rlue Things have finally improved here, with regards to the dependency issues you were having. If you try to implement this again on current master, you should have a much easier time.

rlue:
Thanks guys! It's been a while since I've looked at this code; any thoughts on this proposal?

botandrose:
@rlue Hello! Thank you for taking the time to rebase this and add some tests!

My personal thoughts are that its a bit awkward for this one getter method to return so many different types of values, but I think that's a natural consequence of the configuration API itself being so awkward. So my first inclination is to merge it.

However, there's another consideration: once the dust settles on the v2.0 release, I'd like to pursue a new, non-awkward configuration API (more details in #546), and I imagine that this work might result in us wanting to make breaking changes to this new `#strategy` method, and we would not be able to without releasing v3.0.

So I'm a bit torn. @etagwerker do you have any thoughts?

spec/database_cleaner/configuration_spec.rb
```diff
@@ -174,15 +178,22 @@ def cleaning &block
...
           before do
             config.stub_cleaners [active_record_1,active_record_2,data_mapper_1]
           end
 
+          it "should proxy #strategy to all cleaners and return a single result" do
+            expect(config.strategy)
+              .to eq(active_record_1.strategy)
+              .and eq(active_record_2.strategy)
+              .and eq(data_mapper_1.strategy)
+          end
```
botandrose: I'm a little confused about this assertion. It seems to imply that:
```ruby
config.strategy == active_record_1.strategy &&
  config.strategy == active_record_2.strategy && 
  config.strategy == data_mapper_1.strategy
```
which also seems to imply that all three fake strategies are equal to each other? That doesn't seem quite right. But I might be misunderstanding something here.

botandrose: Ohh, I think I get it. It's saying that the strategy is `:truncation` for all of them, and even though the ORMs differ, the strategy is the same. Hence the `uniq` call in the code. Got it.

rlue:
I felt the same way, but tbh I couldn't figure out a better way, so I just submitted what I had. It does feel like a violation of the robustness principle (specifically, the first half—"be conservative in what you send").

If the intention is to modify it soon in the future, I can just put this PR on hold until the 2.0 release. That way, I won't be increasing the API surface that you guys are responsible for maintaining. (It's not like these changes are desperately important, anyway.)

botandrose:
@rlue okay, cool! I want to make improving the configuration API the next order of business after the v2.0 release is out the door, so if you're cool with postponing this API addition so that it can be a part of that endeavor, I think that could work out pretty well. Let's aim for v2.1!

lib/database_cleaner/configuration.rb
```diff
@@ -13,6 +13,19 @@ def [](orm, opts = {})
       fetch([orm, opts]) { add_cleaner(orm, opts) }
     end 
 
+    def strategy
```
etagwerker: @rlue I think it would be great if you can document this method. It might be unexpected for users that the `strategy` method can return nil, a strategy object, or an array.

etagwerker:
@rlue Now that v2.0 is out, could you please look into this PR? I think it is a good idea but there are now some conflicts with the main branch.

rlue:
Hi folks! Sorry for the delay; hope there is still interest in this PR!

> My personal thoughts are that its a bit awkward for this one getter method to return so many different types of values, but I think that's a natural consequence of the configuration API itself being so awkward. So my first inclination is to merge it.

I have not had the chance to really reflect on this concern, but I think it's valid. If we want to follow the [robustness principle](https://en.wikipedia.org/wiki/Robustness_principle) (that is, to be strict about/consistent in the output that we provide), then rather than supplying potentially three different kinds of output (`nil`, symbol, or hash), we could always return a hash; _i.e.,_ simply

```rb
def strategy
  strategies = transform_values(&:strategy)
end
```

I chose the approach I did based on the assumption that this method would be used primarily for debugging, and thus valued eronomics over consistency. Happy to make amendments as you see fit.

PR #526, opened 2018-04-20
Add an optional URL whitelist safeguard

benclewis:
Hi, love the new safeguards, running DatabaseCleaner has always made me a little nervous.

We run our tests in docker containers so we've had to disable the remote URL check in our pipeline, so I added an option to set a whitelist of URLs.


lib/database_cleaner/safeguard.rb
```diff
@@ -12,8 +12,35 @@ def initialize(env)
...
+      class NotWhitelistedUrl < Error
+        def initialize
+          super("ENV['DATABASE_URL'] is set to a URL that is not on the whitelist. Please refer to https://github.com/DatabaseCleaner/database_cleaner#safeguards")
+        end
+      end
     end
 
+    class WhitelistedUrl
+      def run
+        raise Error::NotWhitelistedUrl if !skip? && given?
```
etagwerker: @benclewis Could you refactor this to avoid the double negative? It's hard to read the combination of `!skip?` and `!DatabaseCleaner.url_whitelist`

Maybe you can make it a little more straightforward like this: 

```
def run
  return if skip?
  raise Error::NotWhitelistedUrl if database_url_not_whitelisted?
end
```

benclewis:
@etagwerker change made, thanks a lot!

etagwerker:
@benclewis Great, thanks! 💯

JuanitoFatas:
@etagwerker Could we cut a new version for this? 🙏

PR #529, opened 2018-04-22
Consider all SQLite DATABASE_URLs local

### LINKED ISSUE #528: New safeguard should consider `sqlite:` urls safe?
Heya, thanks for database cleaner!

I updated to 1.7.0, and the safeguard tripped on this setting:
```
ENV["DATABASE_URL"] = "sqlite3:tmp/test.sqlite3"
```

I'm thinking that perhaps "sqlite3:" urls should always be considered local? AFAICT you can't connect to a remote sqlite server over this procotol. What do you think? I'll submit a PR if this sounds agreeable.

botandrose:
Closes #528


botandrose:
Looks like master is red for the same reason.

PR #530, opened 2018-04-22
Fix travis

botandrose:
Needed to bump Gemfile.lock to be in line with new 1.7.0 release.


kevinburkeomg:
thanks for taking this on!

PR #531, opened 2018-04-22
Bump supported Ruby versions

botandrose:
Ruby 2.1 has been EOL for over a year.

https://www.ruby-lang.org/en/downloads/branches/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,11 @@
 language: ruby
 rvm:
-  - 1.9.3
   - 2.0.0
   - 2.1
   - 2.2
+  - 2.3
+  - 2.4
+  - 2.5
 script:
   - bundle exec rspec
   # - bundle exec cucumber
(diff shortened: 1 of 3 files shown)


botandrose:
Interesting. To support Ruby 2.4 and 2.5, it looks like we'd have to drop support for the old `mysql` gem, which no longer compiles on those versions. This gem hasn't been maintained since 2013: https://rubygems.org/gems/mysql , so dropping support for it is easy to justify, but we'd probably want to major bump the database_cleaner version to v2.0.0. Thoughts?

saiqulhaq:
mysql2 still support ruby 2.0.0
https://github.com/brianmario/mysql2/blob/master/.travis.yml

botandrose:
@saiqulhaq Ah, that's good enough for me, then. I'll restore 2.0.0 and 2.1 to the matrix.

botandrose:
@etagwerker Doing a little digging, I found your comments about splitting up this gem into multiples for database_cleaner v2.0 to restore sanity. I agree something like that is necessary for any progress to be made on this front. Closing.

PR #534, opened 2018-04-23
Use count(*) > 0 instead of count(*)

kevinburkeomg:
count(*) requires searching the entire table, which may be large.
Querying for count(*) > 0 allows MySQL to give up after a single row
is found. Since all we are doing is checking whether the row has
a single table, it should be faster to use count(*) > 0.


kevinburkeomg:
Tests fail for me though it looks like there's an active PR out to fix the build on master, so I'll rebase if/when that gets merged.

ahorek:
according to this article it should be faster
https://www.percona.com/blog/2006/12/01/count-for-innodb-tables/

and your version has a different query plan
```sql
select_type table          type   rows   filtered  Extra
// SELECT COUNT(*) FROM table;
SIMPLE                                             Select tables optimized away
// SELECT COUNT(*) > 0 FROM table;
SIMPLE      table          index 9744244  100      Using index
```

but unfortunatelly I can't confirm it's significantely faster on my setup
MySQL 5.7.17 and Percona 5.7.21-20 / InnoDB
I have only +3% improvement, but in common case DB cleaner doesn't have to clean big tables where using an index could do any difference.
in theory I'm PRO this proposal, because we don't need an exact count.

@kevinburkeomg - do you have your own benchmark? what is your setup?

kevinburkeomg:
I just ran it and I don't see a big improvement either, the bottleneck is probably still the DELETE FROM query. Still it seems like this is an improvement on the status quo.

PR #535, opened 2018-04-24
Fix race conditions in moped and mongo tests

botandrose:
As noted in #522, there is an unstable spec in `spec/database_cleaner/moped/truncation_spec.rb:57` that sometimes fails with:

```
MopedTest::Widget expected to have a count of 0 but was 0
```

This is caused by two race conditions:
1. A race condition in the moped adapter where the truncation command can sometimes return before the truncations are complete.
2. A race condition in the assertion where the count is calculated twice, the first time before the truncations are complete, tripping the assertion, and the second time after the truncations are complete, resulting in the nonsensical error message.

Both have been fixed.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    database_cleaner (1.6.3)
+    database_cleaner (1.7.0)
 
 GEM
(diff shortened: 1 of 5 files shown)


etagwerker:
@botandrose Great, thanks! 👍

PR #537, opened 2018-04-24
Speedup Travis

botandrose:
A few tricks:
1. Run using docker with `sudo: false` for faster startup
2. Cache the result of downloading and extracting neo4j. Even though this is downloading more bytes, this is faster because the cached server is much faster to reach (same network) than the neo4j server.
3. Cache the result of running bundler

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: false
 language: ruby
 rvm:
   - 1.9.3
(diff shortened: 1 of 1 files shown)


botandrose:
Spurious spec failure is unrelated, and fixed by #535

botandrose:
And, we're green! :green_heart:

PR #538, opened 2018-04-25
Fix primary key sequence resetting in Sequel with Postgres and SQLite

botandrose:
There were some pending specs for the Sequel truncation strategy, because it did not reset the primary key sequence after truncation for Postgres and SQLite. Fix these bugs by using the same techniques as used in Datamapper and ActiveRecord strategies, and update the specs to match, greatly simplifying them in the process.


botandrose:
@etagwerker Resolved merge conflict. :green_heart:

PR #541, opened 2018-04-26
Upgrade to Rspec 3

botandrose:
A few notes:

1. We were able to drop a lot of unnecessary dev dependencies by depending on just `rspec`, instead of `rspec-rails`.
2. The upgrade exposed some issues with specs leaking global state modifications. Fixed now.


Guardfile
```diff
@@ -1,9 +0,0 @@
-# A sample Guardfile
-# More info at https://github.com/guard/guard#readme
-
-guard 'rspec', :version => 2 do
-  watch(%r{^spec/.+_spec\.rb$})
-  watch(%r{^lib/(.+)\.rb$})     { |m| "spec/lib/#{m[1]}_spec.rb" }
-  watch('spec/spec_helper.rb')  { "spec" }
-end
-
```
etagwerker: @botandrose No need to remove the `Guardfile`. Please restore it.

database_cleaner.gemspec
```diff
@@ -51,8 +51,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency "couch_potato"
   s.add_development_dependency "sequel", "~> 3.21.0"
   s.add_development_dependency 'ohm', '~> 0.1.3'
-  s.add_development_dependency 'guard-rspec'
```
etagwerker: @botandrose Let's keep `guard-rspec` for now.

spec/database_cleaner/active_record/base_spec.rb
```diff
@@ -1,59 +1,61 @@
...
   end
 end
 
-module DatabaseCleaner
-  describe ActiveRecord do
-    it { should respond_to(:available_strategies) }
-
-    describe "config_file_location" do
-      subject { ActiveRecord.config_file_location }
+describe DatabaseCleaner::ActiveRecord do
```
etagwerker: @botandrose I think the latest convention is to start with `RSpec.describe` instead of `describe` - Could you make it so?

spec/database_cleaner/sequel/deletion_spec.rb
```diff
@@ -1,7 +1,8 @@
-require 'spec_helper'
 require 'database_cleaner/sequel/deletion'
 require 'database_cleaner/shared_strategy'
 require 'sequel'
+require 'support/sequel/sequel_setup'
+# XXX: use ActiveRecord's db_config (`db/config.yml`) for CI/dev convenience
```
etagwerker: @botandrose Interesting. Is it not using it now?

spec/database_cleaner/active_record/base_spec.rb
botandrose: I have set the configuration setting to disable RSpec's monkey patching, entirely, so this convention is enforced, now.

spec/database_cleaner/sequel/deletion_spec.rb
botandrose: Honestly, IDK. I just copied it over along with the line following it from `spec/database_cleaner/sequel/truncation_spec.rb`. I think its intended as an FYI, but maybe its an outdated TODO, and we can just ditch it. I'll git blame it, and find out.

botandrose: Ok, looks like its an advisory comment. I think its just saying something like "despite the fact that we're testing the Sequel ORM, let's just cheat and reuse the ActiveRecord connection .yml file, because its basically free."

database_cleaner.gemspec
botandrose: Oh, yeah, I remember what happened now. Bumping `rspec` required bumping `guard-rspec` which required bumping `guard` which requires Ruby 2.2+. Ugh. I'm going to see if there's some version in the middle that will work.

PR #550, opened 2018-05-04
Update cucumber step defs to RSpec 3 assertion syntax

botandrose:
Fixes a bunch of warnings. Missed this in the RSpec 3 update PR.

### DIFF
--- a/examples/features/step_definitions/activerecord_steps.rb
+++ b/examples/features/step_definitions/activerecord_steps.rb
@@ -11,7 +11,7 @@ When /^I create a widget using activerecord$/ do
 end
 
 Then /^I should see ([\d]+) widget using activerecord$/ do |widget_count|
-  ActiveRecordWidget.count.should == widget_count.to_i
+  expect(ActiveRecordWidget.count).to eq widget_count.to_i
 end
 
 When /^I create a widget in one db using activerecord$/ do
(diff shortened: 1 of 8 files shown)


mauro-oto:
@botandrose seems like there's a flaky test that was being kept at bay by some code removed by the RSpec 3 PR here: https://github.com/botandrose/database_cleaner/blob/f3feaa36ba463a558ae7c115a7ebe60ea08a5e7f/spec/database_cleaner/moped/truncation_spec.rb#L27-L34 

The failure:
```
1) DatabaseCleaner::Moped::Truncation by default truncates all collections
     Failure/Error:
       expect { subject.clean }.to change {
         [MopedTest::Widget.count, MopedTest::Gadget.count, MopedTest::System.count]
       }.from([1,1,1]).to([0,0,0])
     
       expected `[MopedTest::Widget.count, MopedTest::Gadget.count, MopedTest::System.count]` to have changed from [1, 1, 1] to [0, 0, 0], but did not change
     # ./spec/database_cleaner/moped/truncation_spec.rb:25:in `block (3 levels) in <top (required)>'
     # ./spec/database_cleaner/moped/truncation_spec.rb:11:in `block (2 levels) in <top (required)>'
```

I just restarted the build, but we should probably re-introduce that patch or something similar to it.

botandrose:
@mauro-oto Thanks for taking a look at this. However, I think this failure is something different.

The removed `ensure_counts` code functioned as a sanity check to assert that the before and after state were as expected. I replaced it with `expect {}.to change {}.from().to()` which also asserts the before and after state in the `from` and `to` clauses. Basically replacing an ad-hoc assertion with a more expressive one that is built-in to RSpec, so the assertion is still intact... nothing was removed.

I think this flaky failure is something new. Probably a race condition between MongoDB and the test process (MongoDB is loves to return before a posted update has actually completed). I'm going to try to reproduce the failure locally, and will post a new PR with a fix when successful.

PR #552, opened 2018-05-05
Refactor specs

botandrose:
Looking at the test suite, I've noticed two related issues that could be improved:

1. There is a ton of code centered around resetting the database state after tests (which is fairly ironic for this project, when you think about it). There is a lot of duplication here, as each orm and db test suite have their own code for dealing with this.
2. Some orm and db combos are exhaustively tested, e.g. AR, while some are barely tested at all, e.g. DM. Both AR and DM have virtually identical functionality with regards to DatabaseCleaner, so we should be able to reuse the exhaustive AR specs for DM and other orm and db combos, but we are currently not doing very much of this.

This PR is a first stab at improving this situation, mostly by addressing issue 1, for now. The goal is to abstract away the reset details behind a test helper interface that can be reused for all orm and db combos, thus unlocking more shared test scenarios to address issue 2. This will ultimately help the DatabaseCleaner 2.0 effort, by allowing external adapter gems to pull in much more shared scenarios from core.

I stopped short at refactoring the SequelHelper, because of pending PR #538, which heavily modifies the Sequel specs, so the merge conflict would be awful. Waiting for that one to merge, first.


lib/database_cleaner/null_strategy.rb
```diff
@@ -1,18 +1,18 @@
 module DatabaseCleaner
   class NullStrategy
-    def self.start
+    def start
```
ahorek: this is an API change
```
DatabaseCleaner.strategy = DatabaseCleaner::NullStrategy
```
woun't work anymore, right?

botandrose: Ah, I assumed that the null strategy was only used internally. Upon reflection, that is obviously not a safe assumption! This change was tangential to the main thrust of the PR, anyways, so I'll just kill that particular commit. Thanks for catching this!

PR #557, opened 2018-05-21
Only test via public APIs, and don't mock or stub SUT

botandrose:
I started investigative work on extracting adapter gems, and almost immediately ran into issues with broken tests, because the tests were tightly coupled with the implementation details. These issues generally surfaced in two forms: testing private methods instead of testing behavior via the public API, mocking and stubbing internal methods directly on the system under test instead of the system's collaborators.

This PR addresses those issues, opening the door for friction-less refactoring and feature addition. Hopefully this is the last yak-shave before we can start delivering real value i.e. adapter gems!


spec/database_cleaner/active_record/truncation_spec.rb
```diff
@@ -1,113 +1,96 @@
...
+            subject.clean
+            expect(User.create.id).to eq 1
+          end
 
-        it "should raise an error when :only and :except options are used" do
-          expect {
-            Truncation.new(:except => ['widgets'], :only => ['widgets'])
-          }.to raise_error(ArgumentError)
-        end
+          xit "should not reset AUTO_INCREMENT index of table if :reset_ids is false" do
```
etagwerker: @botandrose Why was this one failing? Maybe you can create an issue for this so that we don't forget about it.

botandrose: Good call. This was the result of extending an existing test to cover more dbs. After doing so, it turns out that this doesn't work for some reason on some dbs. New information! I will create an issue

botandrose: Whoops, I was mistaken! I got this mixed up with another spec that had the situation I described! In any event, I have created an issue for this: #559

botandrose:
@etagwerker Okay, just resolved the merge conflict, so this should be ready to go! :green_heart:

After this is merged, I can push my gem extraction branch up for your review. It's happening! :tada:

PR #560, opened 2018-05-28
Extract adapter gems

botandrose:
Okay, I've got something working here! I've opted to extract the gems into the `/adapters` dir in this repo, to keep all the changes cohesive for this PR. I'm thinking that we can punt on extracting them to their own separate repos until right before we're ready to release... there are a few more changes I'd like to make in follow-up PRs.

Other items of note:
1. This is backwards compatible! I've set the versions of core and adapters to 1.8.0. This is a big win, I think, because we can provide both the deprecated behavior and the supported behavior in one release. This should grealy improve developer UX, and allow for 2.0 to just be a "garbage collection" release where we only remove deprecated APIs, which I think is the best way to deal with breaking changes. Keeping the gem at 1.x also allows us to punt on decision-making for other breaking changes.

2. Running `rake` tests both the core gem and the adapter gems, for the moment. This means Travis is testing everything, too.

3. The core gem's cucumber suite now tests both scenarios: the deprecated built-in adapters, and including the new adapter gems, to make sure both work. The deprecation warnings are visible in the output of the former.

4. READMEs for adapter gems still need to be edited.

What do you think?


botandrose:
Oops, just noticed that the `bundle gem` command I ran to generate the adapter gem skeletons put my name in all the MIT licenses! I'm not trying to take over. :)

Looking at the core LICENSE, it's © 2009 Ben Mabey. Should the gem LICENSEs match that?

botandrose:
@etagwerker Did you get a chance to take a gander at this? What do you think?

botandrose:
@etagwerker Anything holding this back from merge?

botandrose:
@etagwerker Okay, rebase wasn't too bad. Good to go, I think!

This did highlight a downside to this 1.x strategy, however. Until 2.0, future PRs will have to make their changes twice, once in the base gem, and again in the adapter gem, if the changes are db-specific.

Still worth it, overall, IMO, but something to be aware of until we split the gems out completely in 2.0.

Anything else I can do here?

botandrose:
@etagwerker It occurs to me that we might be able to reduce this duplication with symlinks. But I have th following concerns:

1. Will all contributors understand symlinks?
2. Windows doesn't support symlinks, I think?

What do you think?

etagwerker:
@botandrose I see what you did and I'm wondering if we could remove duplication without symlinks. I don't think using symlinks is a good idea. I'm afraid of the issues it might cause in Windows environments (I'm on a Mac) 

Would it be possible to require the ORM-specific files from the adapters folder and see how that does? I don't like the idea of merging this and ending up with so much duplication. 

I wonder if you could change `lib/database_cleaner.rb` to require all the files from the `adapters` folder.

botandrose:
@etagwerker Hey, that's a great idea! Much simpler than symlinks. I've set it up to add the adapter gem's lib directory to the `$LOAD_PATH` when the adapter is being autoloaded, and it works great! Big red diff now! <3

I can't think of a reason not to merge this, anymore. Can you think of anything?

botandrose:
@etagwerker Happy New Year! Did you get a chance to take a look at this? I'd love to do some more follow-up PRs to prepare for the 1.8.0 release!

botandrose:
@etagwerker Ping

botandrose:
@etagwerker Hello! Just checking in again to see if you have found any time to give this a review. Thank you!

botandrose:
@etagwerker Just rebased to resolve the merge conflict introduced by the PR merged today.

etagwerker:
Great, thanks! I'll review this PR this week. Sorry for the delay!

etagwerker:
@botandrose Looks good, just merged it! 

One thing we should discuss is the way documentation is going to look like. I think that we should just document usage in `database_cleaner` and later include a link to it from the adapter gems -- what do you think?

botandrose:
Hooray! A big step forward! Now we can finally start doing Appraisals for each of the adapter gems and actually test them with current versions of their dependencies! I'll make a PR for this, and another for documentation. Is there anything else you'd like to see before we cut a 1.8.0 release?

etagwerker:
@botandrose I wonder if we should cut a 1.8.0 or a 1.99.0 release. 1.99.0 would be a way to signal that 2.0 is coming and that they will need to change some code in their application for `database_cleaner` to work again.

Before we cut a new release, I'd like to make sure that we are not polluting the test.log with a lot of deprecation warnings. When I ran the test suite for this gem I got a ton of deprecation messages. I haven't tested a sample app with `master` so I'd like to do that and make sure it is not too noisy.

botandrose:
@etagwerker Good call on the chatty deprecation warnings. I'll open a PR to make sure each deprecation warning is only printed once.

With regards to the question of 1.8.0 vs 1.99.0, either is fine by me. I went with 1.8.0 because that more closely follows semver, but if you have a preference to 1.99.0, that's cool with me, too. Just say the word and I'll bump the versions.

Issue #561, opened 2018-06-11
Unix sockets should be considered local with Postgres


fdr:
Similar to #528, localish URLs relying on unix sockets are miscategorized with Postgres.

Unix socket urls can look like: `postgres:///dbname` or `postgresql:///dbname` (when using the default unix socket baked into the linked `libpq`) or, more flexibly, `host=/tmp` (`libpq` detects unix sockets with a leading slash: only absolute paths are supported).

I think there is also an IPv6 (`::1`) omission.

botandrose:
Looks like this was addressed by #569. Reopen if not.

PR #563, opened 2018-07-12
Unix sockets should be considered local with Postgres

faucct:
https://github.com/DatabaseCleaner/database_cleaner/issues/561

### DIFF
--- a/lib/database_cleaner/safeguard.rb
+++ b/lib/database_cleaner/safeguard.rb
@@ -39,8 +39,6 @@ module DatabaseCleaner
 
 
     class RemoteDatabaseUrl
-      LOCAL = %w(localhost .local 127.0.0.1 sqlite3:)
-
       def run
         raise Error::RemoteDatabaseUrl if !skip? && given?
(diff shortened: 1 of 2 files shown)


faucct:
ping

lib/database_cleaner/safeguard.rb
```diff
@@ -52,7 +50,13 @@ def given?
         end
 
         def remote?(url)
-          url && !LOCAL.any? { |str| url.include?(str) }
+          url &&
+            case URI(url).host
+            when nil, 'localhost', /\.local/, '127.0.0.1'
```
etagwerker: @faucct Could you keep this as a constant? 

Also, it seems that you missed `sqlite3:` -- was that on purpose?

faucct: ping

ivanbrennan:
It looks like this has been addressed in a different PR: https://github.com/DatabaseCleaner/database_cleaner/pull/569

PR #567, opened 2018-10-05
Refactor MySQL deletion strategy (AR)

### LINKED ISSUE #516: warning: instance variable @table_stats_query not initialized
Always I run the test with `:deletion` strategy, I got this error:

```
database_cleaner-1.6.2/lib/database_cleaner/active_record/deletion.rb:66: warning: instance variable @table_stats_query not initialized
```

My environment:

- OSX 10.13.2 (17C88)
- ruby 2.4.2p198
- rspec (3.7.0)
- mysql2 (0.4.10)
- database_cleaner (1.6.2)
- activerecord (5.1.4)

Workaround:

As workaround, I fixed as follow:

```
if @cache_tables && defined?(@table_stats_query) && !@table_stats_query.nil?
  return @table_stats_query
else
```

buehmann:
While trying to understand the special deletion strategy for MySQL, I refactored code and DB queries. Fixes #516 as well.

For the future: There appears to be some overlap with the truncation strategy. What is effectively done here is a `pre_count` for deletion.

### DIFF
--- a/lib/database_cleaner/active_record/deletion.rb
+++ b/lib/database_cleaner/active_record/deletion.rb
@@ -56,7 +56,7 @@ module DatabaseCleaner::ActiveRecord
       @db_name ||= connection.instance_variable_get('@config')[:database]
       stats = table_stats_query(connection, @db_name)
       if stats != ''
-        connection.exec_query(stats).inject([]) {|all, stat| all << stat['table_name'] if stat['has_rows'] == 1; all }
+        connection.select_values(stats)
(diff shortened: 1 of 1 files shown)


ahorek:
LGTM, cc @etagwerker 

btw I tested UNION vs UNiON ALL performance. It's not faster according to my benchmark, but it make sence to use it here because table names are already unique

etagwerker:
@buehmann Thanks! 👍

PR #568, opened 2018-11-05
Improve "skimmability" of additional truncation options README description 

MarkyMarkMcDonald:
There aren't any specific examples of setting these options, and it took us a bit
to understand they were similar to the options above.

Signed-off-by: Salman Hossain <salman.hossain@greenhouse.io>

CC @salmanhin


JonRowe:
Improves the readability :+1:

PR #569, opened 2018-11-07
Allow postgres:///dbname as a local url

will:
Hi, I updated today and was surprised (and a little worried at first) that I was about to clean a remote database. It turns out I was using a style of DATABASE_URL that wasn't considered local. This patch adds that, and also tightens down the checks to actually look at the host (or protocol) instead of those appearing somewhere in the string.

### DIFF
--- a/lib/database_cleaner/safeguard.rb
+++ b/lib/database_cleaner/safeguard.rb
@@ -39,7 +39,7 @@ module DatabaseCleaner
 
 
     class RemoteDatabaseUrl
-      LOCAL = %w(localhost .local 127.0.0.1 sqlite3:)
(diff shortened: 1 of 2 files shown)


will:
@botandrose I was also considering not having it be a constant at all, but turning on `# frozen_string_literal: true` for this file. I think that'd be goodd, but I thought maybe that was over-reaching for this PR.

etagwerker:
> @botandrose I was also considering not having it be a constant at all, but turning on # frozen_string_literal: true for this file. I think that'd be goodd, but I thought maybe that was over-reaching for this PR.

@will Sounds like a good idea. If you submit a new PR for that, I'd really appreciate it. Thanks!

PR #574, opened 2019-04-04
Fix MySQL deprecation warnings with Rails 5

stanhu:
When the truncation strategy is used with Rails 5, we see
the deprecation message:

```
DEPRECATION WARNING: #tables currently returns both tables and
views. This behavior is deprecated and will be changed with Rails 5.1 to
only return tables. Use #data_sources instead.
```


stanhu:
@etagwerker Would you mind reviewing this? This deprecation warning is polluting our MySQL CI logs like crazy. :)

PR #578, opened 2019-05-14
 Add SQLServerAdapter with view_cache override method

ablignaut:
I created a SQLServerAdapter to so I could override the implementation of the database_cleaner_view_cache method since sql server.

The other option was to change the underlying implementation of https://github.com/DatabaseCleaner/database_cleaner/blame/master/lib/database_cleaner/active_record/truncation.rb#L25 
I'm not sure why this is part of the base implementation?
```ruby
 where table_schema = '#{current_database}'
```

### DIFF
--- a/VERSION.yml
+++ b/VERSION.yml
@@ -1,4 +1,4 @@
 ---
 :major: 1
 :minor: 7
-:patch: 0
+:patch: 1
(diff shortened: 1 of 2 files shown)


botandrose:
@ablignaut Thank you for the PR! However, the ActiveRecord adapter has been extracted into its own repo at https://github.com/DatabaseCleaner/database_cleaner-active_record. If you close this PR and reopen it over there, I'd be happy to take a look!

PR #579, opened 2019-05-20
Refactoring and deprecation

botandrose:
Now that we've extracted the adapter gems, its time to start making life a little bit better for ourselves, by cleaning up internals, and paving the way for 2.0 with deprecations. Turns out I had this branch all ready to go!

Notable changes:
* Extract autodetection bits into `DatabaseCleaner::ORMAutodetector` and deprecate them for easy removal in 2.0
* Deprecate unnecessarily-public methods that were only used internally so we can privatize them in 2.0
* Rename `DatabaseCleaner.connections` to the more accurate `DatabaseCleaner.cleaners` and deprecate the old method
* Extract `DatabaseCleaner::Cleaners` container object to encapsulate knowledge of currently loaded cleaners
* Misc refactoring

### DIFF
--- a/lib/database_cleaner/configuration.rb
+++ b/lib/database_cleaner/configuration.rb
@@ -2,21 +2,20 @@ require 'database_cleaner/base'
 
 module DatabaseCleaner
 
-  class NoORMDetected < StandardError;   end
-  class UnknownStrategySpecified < ArgumentError;   end
+  class NoORMDetected < StandardError; end
+  class UnknownStrategySpecified < ArgumentError; end
 
   class Configuration
-    def [](orm,opts = {})
+    def initialize
+      @cleaners ||= {}
+    end
+
+    # FIXME this method conflates creation with lookup... both a command and a query. yuck.
+    def [](orm, opts = {})
       raise NoORMDetected unless orm
-      init_cleaners
-      # TODO: deprecate
-      # this method conflates creation with lookup.  Both a command and a query. Yuck.
-      if @cleaners.has_key? [orm, opts]
-        @cleaners[[orm, opts]]
-      else
-        add_cleaner(orm, opts)
-      end
+      @cleaners.fetch([orm, opts]) { add_cleaner(orm, opts) }
     end 
+
     attr_accessor :app_root, :logger
 
     def app_root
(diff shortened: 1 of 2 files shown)


botandrose:
@etagwerker Any idea why Travis isn't testing PRs anymore? I noticed it when I force-pushed my last PR, but thought it might be a fluke. It appears that it isn't...

botandrose:
@etagwerker https://travis-ci.org/DatabaseCleaner/database_cleaner says that the repo is disabled. Do you know why?

etagwerker:
@botandrose No idea why that happened. I just re-activated it. 👍

botandrose:
@etagwerker Okay, pushed another commit to change the deprecation warnings to only print once per deprecated method.

botandrose:
@etagwerker Did you get a chance to take a look at this?

etagwerker:
@botandrose Sorry for the delay. It would be very helpful if you split this PR into smaller pull requests. 

Could you submit one PR for each of these? 

- Extract autodetection bits into DatabaseCleaner::ORMAutodetector and deprecate them for easy removal in 2.0

- Deprecate unnecessarily-public methods that were only used internally so we can privatize them in 2.0

- Rename DatabaseCleaner.connections to the more accurate DatabaseCleaner.cleaners and deprecate the old method

- Extract DatabaseCleaner::Cleaners container object to encapsulate knowledge of currently loaded cleaners

botandrose:
@etagwerker Hmm, in theory, yes. The challenge is that most of the commits in this PR are built on top of the ones before it. I could submit one PR for the first commit, and then after that gets merged or not in one way or another, rebase the remaining commits and submit the new first one in a new PR, rinse and repeat?

etagwerker:
@botandrose Sounds good. Let's start with the first one then. 👍

botandrose:
@etagwerker Okay, I've force pushed this branch to just include the first commit, which focuses on cleaning up how the `Configuration` object manages its internal state. There is one deprecation: the `#init_cleaners` method, which I sincerely doubt anyone was ever calling, but it was a public method, so the right thing to do is deprecate it before removing it, I think.

botandrose:
@etagwerker Looks like Travis is failing. I'll take a stab at that, as well.

botandrose:
@etagwerker Okay, rebased on master, so this is green now, ready for your review.

PR #580, opened 2019-05-20
Update documentation for new adapter gems

botandrose:
I started this PR to try to get documentation in shape for the upcoming 1.8.0 release. After trying out a few different ways of organizing the information, it seems to me like the easiest way to go about this is for the base README to cover general usage of the library, while the adapter gems' READMEs cover configuration and gotchas specific to the adapter.

This PR is a Work In Progress, because I still need to go through and work all the READMEs for the various adapters. However, I think the base README and the ActiveRecord adapter README look pretty good. What do you think? If you like this direction, I'll do the work to finish up the adapter READMEs.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -32,8 +32,6 @@ end
 
 ## Supported Databases, Libraries and Strategies
 
-ActiveRecord, DataMapper, Sequel, MongoMapper, Mongoid, CouchPotato, Ohm and Redis are supported.
-
 Here is an overview of the strategies supported for each ORM:
 
 Gem | ORM | Truncation | Transaction | Deletion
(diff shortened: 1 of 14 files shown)


etagwerker:
@botandrose Hey, I haven't reviewed the PR yet but I wanted to say that I like the direction. I was thinking the same thing: Document the main usage examples in `database_cleaner` and move adapter-specific documentation to the adapter's README. 👍 

Thanks!

botandrose:
@etagwerker Okay, while its not perfect, I think this puts the documentation in a good enough state where we could release it without too much embarassment. What do you think?

botandrose:
@etagwerker Did you get a chance to take a look at this?

etagwerker:
@botandrose Looks good, thanks!

PR #581, opened 2019-05-21
Fix Travis

botandrose:
Get Travis green again. Had to downgrade OS so that Neo4j would run on it, and fix a small error in the couch potato adapter bin/setup script.


botandrose:
:green_heart:

PR #585, opened 2019-08-26
Fix broken links

luotaoruby:


etagwerker:
@luotaoruby Thanks for this! I prefer relative links for now, so I'm going with #592

PR #589, opened 2019-09-10
Update master docs to link to the current version README and docs

borenho:
Add a note and link to the `master` branch README to lead to the latest version README


README.markdown
```diff
@@ -8,6 +8,9 @@ Database Cleaner is a set of gems containing strategies for cleaning your databa
 The original use case was to ensure a clean state during tests.
 Each strategy is a small amount of code but is code that is usually needed in any ruby app that is testing with a database.
 
+**NB** Please note that this README is not of the gem's latest version, you may
```
JonRowe: This should say `this README is for the unreleased development version` as its technically the latest version, just not the last release.

README.markdown
```diff
@@ -8,6 +8,9 @@ Database Cleaner is a set of gems containing strategies for cleaning your databa
...
+find some unexpected errors if you go by it. Find the current version README and docs here: [most recent docs](https://github.com/DatabaseCleaner/database_cleaner/tree/v1.7.0)
```
JonRowe: Rather than unexpected errors, its better to say something like `you will find inconsistencies compared to 1.x` they're not errors, they just don't apply to 1.7. You could link to the read me directly here, rather than the root.

PR #595, opened 2019-09-23
Fix travis

### LINKED ISSUE #593: Travis CI build is broken in `master`
Travis CI configuration is out of date and builds are no longer working. For example: https://travis-ci.org/DatabaseCleaner/database_cleaner/jobs/588443682

botandrose:
Looks like the Gemfile.lock was missing a dependency? Not sure how this could have possibly happened, but whatever. Also locked down the bundler version in Travis config to keep things reproducible. Fixes #593

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ rvm:
   - 2.1
   - 2.2
 before_install:
+  - gem install bundler -v 1.17.3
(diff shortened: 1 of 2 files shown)


botandrose:
@etagwerker :green_heart: :tada:

PR #601, opened 2019-12-10
Introduce DatabaseCleaner.cleaners as a replacement for DatabaseCleaner.connections

botandrose:
Back in the Ruby 1.8 days, hashes didn't guarantee insertion order, so an array was used internally to store the collection of cleaners. This has since been refactored to use a hash, which effectively matches the configuration API, e.g.: `DatabaseCleaner[:mongoid]`, but the old array accessor method remains in the public API. Therefore, I think we should deprecate the old array accessor for removal in v2.0, and replace it with a new getter `.cleaners` which exposes the internal hash configuration in a way that matches the configuration API.

### DIFF
--- a/lib/database_cleaner.rb
+++ b/lib/database_cleaner.rb
@@ -11,6 +11,8 @@ module DatabaseCleaner
       :app_root,
       :logger=,
       :logger,
+      :cleaners,
+      :cleaners=,
       :strategy=,
       :orm=,
       :start,
--- a/lib/database_cleaner/configuration.rb
+++ b/lib/database_cleaner/configuration.rb
@@ -16,7 +16,7 @@ module DatabaseCleaner
       @cleaners.fetch([orm, opts]) { add_cleaner(orm, opts) }
     end 
 
-    attr_accessor :app_root, :logger
+    attr_accessor :app_root, :logger, :cleaners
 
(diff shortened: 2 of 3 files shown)


lib/database_cleaner.rb
```diff
@@ -11,6 +11,8 @@ class << self
       :app_root,
       :logger=,
       :logger,
+      :cleaners,
+      :cleaners=,
```
etagwerker: @botandrose Isn't this a non-backwards compatible change? What happens if someone is using `DatabaseCleaner.cleaners` in their code? They might get a `NoMethodError`, right?

botandrose: @etagwerker Hmm, I'm not understanding what you mean? This PR _adds_ the `DatabaseCleaners.cleaners` method. It also deprecates `DatabaseCleaners.connections` but doesn't remove it yet, so it should be Semver forwards compatible. Can you explain more about your concern? Maybe I am misunderstanding something. :)

etagwerker: @botandrose Oh, good point! I misread the set of changes. Sorry about that. 👍

botandrose: All good! Thank you for the review! 😊

PR #603, opened 2019-12-11
Tighten up Public API

botandrose:
DatabaseCleaner's public API contains many methods that are just used for internal bookkeeping or implementation, and should have been private methods. Since privatizing a method is a breaking change, we need to deprecate usage of these methods before privatizing or removing them in v2.0. I don't expect anyone is actually using this methods, because they're undocumented, and not of much utility on their own, but its possible.

There's also a few methods that are old undocumented aliases for other methods, so let's deprecate those, too, while we're at it!


botandrose:
No dissent, so merging!

PR #620, opened 2020-02-01
v2.0.0.beta

botandrose:
Oh boy, here it is! This branch contains all the big changes needed for the v2.0.0.beta release:

1. Removal of all adapters
2. Removal of all deprecated code
3. Drop support for EOL Rubies and add support for current Rubies
4. Rename gem to `database_cleaner-core`
5. Add new `database_cleaner` metagem that depends on `database_cleaner-active_record`

Not ready to merge this yet... I want to get all the adapter repos into a good spot first. But I figured its good to get all this out in the open nice and early, so that people can take a look and give any feedback they might have!

Exciting times for database_cleaner!


botandrose:
@etagwerker @JonRowe All right! I think this PR + https://github.com/DatabaseCleaner/database_cleaner-active_record/pull/3 represent the current vision for v2.0.0. Will y'all take a look and let me know what you think?

When we agree that this all looks good, I'll finish up the other adapters and release v2.0.0.beta versions of everything for wider testing.

Very exciting to finally escape from dependency hell!

spec/database_cleaner/base_spec.rb
```diff
@@ -231,23 +95,29 @@ module DatabaseCleaner
     end
 
     describe "clean_with" do
-      # FIXME hacky null strategy
-      # because you can't pass a NullStrategy to #clean_with
+      subject { described_class.new(:active_record) }
```
JonRowe: I know this to make the existing specs work, but if you're modernising here, we, the rspec team, recommend you at least use named `subject`, and we don't really recommend using `described_class` (as it requires constants to be loaded before use) but that might be something for a later PR.

Gemfile.lock
```diff
@@ -1,153 +1,20 @@
 PATH
   remote: .
   specs:
-    database_cleaner (1.8.2)
```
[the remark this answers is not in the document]
botandrose: Can you explain the value in gitignoring Gemfile.lock in library repos? I've never been entirely clear on that. I'm guessing its that future test runs will pick up newly released versions of dependencies, giving us a heads-up if one of them fails?

spec/database_cleaner/configuration_spec.rb
```diff
@@ -93,6 +86,9 @@ def stub_cleaners(array)
     end
 
     it "should retrieve a db rather than create a new one" do
+      stub_const "DatabaseCleaner::ActiveRecord", Module.new
+      stub_const "DatabaseCleaner::ActiveRecord::Truncation", Class.new
```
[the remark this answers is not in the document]
botandrose: Hmm, I having a hard time getting this to work with `object_double`, but I may be doing it wrong! Can you say a bit more about how this might work?

README.markdown
```diff
@@ -1,9 +1,8 @@
-<!--
 **If you're viewing this at https://github.com/DatabaseCleaner/database_cleaner,
 you're reading the documentation for the `master` branch.
 [View documentation for the latest release
-(1.7.0).](https://github.com/DatabaseCleaner/database_cleaner/blob/v1.7.0/README.markdown)**
--->
+(1.8.1).](https://github.com/DatabaseCleaner/database_cleaner/blob/v1.8.1/README.markdown)**
```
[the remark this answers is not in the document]
botandrose: Now that you've pointed this out, I'm actually thinking this should be v1.8.2... until v2.0.0 proper lands (not pre-release), we run the risk of confusing normal users of database_cleaner with documentation for newer versions. Once v2.0.0 lands we can kill this whole paragraph.

lib/database_cleaner-core.rb
```diff
@@ -0,0 +1 @@
+# empty file to avoid LoadError if someone requires database_cleaner-core directly
```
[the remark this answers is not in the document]
botandrose: Yeah, I'm not really sure what the best thing to do is here. I can't think of a single use-case for intentionally doing a `require "database_cleaner-core"`. The only thing I could come up with is someone wanting to pin core down to a specific version for some reason in their Gemfile, and then not specifying `require: false`, in which case that require would be issued when `Bundler.require` happens. Thus this empty file. Shrug?

spec/database_cleaner/base_spec.rb
botandrose: TIL about named subjects! :smiley: I'll give that a go.

Also, if you ever want to make a pass through the test suites, they all need a lot of work! A decade of being tied to the 2010-era gem ecosystem, while the world moves on, has necessarily yielded a paradigm of testing implementation, not behavior. :/ I've already done some work on the DC::AR adapter gem's suite, but once this PR lands, we're finally free to modernize things!

lib/database_cleaner-core.rb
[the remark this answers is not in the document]
botandrose: Hmm, thinking about this even more, I think this current strategy is not going to work as-is.

The goal behind the `database_cleaner-core` and `database_cleaner` metagem is to reduce v1 - v2 breakage for the majority of database_cleaner's consumers, i.e. Rails apps. By making `database_cleaner` a metagem for `database_cleaner-active_record`, ,ost Rails apps should be able to upgrade from 1.7 to 1.8 to 2.0 with no deprecation warnings or changes.

However, I'm just now realizing that with this strategy `Bundler.require` on a Gemfile with `gem "database_cleaner"` will just load core without loading the AR adapter, even though its pulled in as a dependency. This whole strategy might need to be rethought...

spec/database_cleaner/base_spec.rb
botandrose: Okay, I've pushed a new commit that names all custom subjects, and removes usage of `described_class`, as well as a number of other cleanups. WDYT?

Gemfile.lock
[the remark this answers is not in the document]
botandrose: Personally, I've always weighed this against build/failure reproducibility across development instances and found the latter more compelling, but I'd love to learn something new here. What are your views?

lib/database_cleaner-core.rb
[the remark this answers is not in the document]
botandrose: I'm thinking the solution could be:
* lib/database_cleaner-core.rb: `require "database_cleaner/core"`
* Move existing lib/database_cleaner.rb to lib/database_cleaner/core.rb
* New lib/database_cleaner.rb: `require "database_cleaner/active_record"`

I think this checks all the boxes?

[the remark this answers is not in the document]
JonRowe: That sounds sensible

database_cleaner-core.gemspec
```diff
@@ -0,0 +1,30 @@
...
+  spec.require_paths = ["lib"]
+
+  spec.add_development_dependency "rake"
+  spec.add_development_dependency "bundler"
+  spec.add_development_dependency 'guard-rspec'
+  spec.add_development_dependency "listen"
+  spec.add_development_dependency "rspec"
+  spec.add_development_dependency "cucumber"
+end
+
```
JonRowe: Random empty line?

Gemfile.lock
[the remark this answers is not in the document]
JonRowe: It prevents builds from staying in sync with reality. Your gain stability in your builds by locking to a period of time rather than versions. If you have specific versions that you need for a build to pass they should be in your `gemspec`, or if you need some custom magic in your `Gemfile`.

Someone who consumes the library will not get the versions specified in your `Gemfile` or `Gemfile.lock` no matter how they get the library, so you end up not seeing failures that affect them until someone bothers to open an issue.

README.markdown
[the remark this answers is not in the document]
JonRowe: Ah I see, yes.

spec/database_cleaner/base_spec.rb
```diff
@@ -223,31 +90,19 @@ module DatabaseCleaner
...
       before do
-        allow(::ActiveRecord::Base).to receive(:connection).and_return(double.as_null_object)
-        allow(strategy_class).to receive(:new).and_return(strategy)
+        stub_const "DatabaseCleaner::ActiveRecord::Truncation", strategy_class
       end
 
       it "should pass all arguments to strategy initializer" do
         expect(strategy_class).to receive(:new).with(:dollar, :amet, ipsum: "random").and_return(strategy)
+        expect(strategy).to receive(:clean)
         subject.clean_with :truncation, :dollar, :amet, ipsum: "random"
```
JonRowe: I'd probably use the named subject here for example.

Gemfile.lock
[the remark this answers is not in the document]
botandrose: @JonRowe Ah, that makes sense, and its a particularly relevant argument, considering the amount of effort this library is currently expending to address exactly that situation, i.e. tied to a 2010 reality!

I wonder though, does the existence of dependabot completely change this situation? Now that its on the job, do we get the best of both worlds by leaving Gemfile.lock checked in? Developers of the library don't suffer from works-on-my-machine, and dependabot ensures that we stay in sync with reality?

database_cleaner-core.gemspec
botandrose: Ah, I have a habit of ending all my files with carriage returns to make diffs cleaner when lines are added to the end. This file will probably never have that happen though, so I'll remove.

JonRowe: This is two carriage returns, the posix standard has all lines end with `\n` where as this is `\n\n` or it wouldn't show up in github. Lines lacking a `\n` are equally invalid at the end of a file and are what cause the Github red symbol. A properly terminated file won't have any more diff than the new line you're adding.

spec/database_cleaner/base_spec.rb
botandrose: Named subject is a brand new concept to me, so I very much value your input, here. Thank you!

I've already named the subject `:active_record_cleaner` in the `describe "clean_with"` context, with the idea that it aids in debugging via better failure messages. But you're saying that rather than invoking that named subject as `subject`, it should be invoked as `active_record_cleaner`? And now I'm realizing that named subject is primarily a readability win? Does this sound right?

lib/database_cleaner-core.rb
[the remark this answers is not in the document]
botandrose: Okay, just pushed a new commit resolving this.

database_cleaner-core.gemspec
botandrose: Wow, TIL! Awesome.

spec/database_cleaner/base_spec.rb
JonRowe: Yep it should be invoked as its name, subject is a left over from when we recommended one line specs, it was never intended to be invoked by its name, but then it became a symbol of the "subject under test" when really its clearer if this is named to what it is, named subjects are a good cross as it highlights which helper is the subject under test, but encourages better naming.

lib/database_cleaner/core.rb
```diff
@@ -0,0 +1,30 @@
...
+    attr_accessor :allow_remote_database_url, :allow_production, :url_whitelist
+
+    private
+
+    def configuration
+      @configuration ||= Configuration.new
+    end
+  end
+end
+
```
JonRowe: 🙃

spec/database_cleaner/base_spec.rb
botandrose: I dig it, man. Thank you for the explanation! I'll update the specs with this new understanding.

lib/database_cleaner/core.rb
botandrose: LOL its gonna be hard to break this habit!

spec/database_cleaner/base_spec.rb
botandrose: One last question about this. If it were you, would you remove all invocations of `subject` in it blocks? i.e. the only reason to use `subject` in your specs is to set up a named subject? Basically a version of `let` that carries an "primary SUT" connotation?

I'm not sure where to draw the line, right now. I'm waffling between the path described above, and using bare `subject` when all we need is an implicit `described_class.new`, and named subjects otherwise.

What would you do?

JonRowe: > Basically a version of let that carries an "primary SUT" connotation?

They are a `let` under the hood ;) Yes I would, but I don't even use `subject` in my own tests.

botandrose: Okay, I've redone all the specs in this style, and I agree that it is an improvement in readability! Thank you for helping me level up my RSpec skills!

spec/database_cleaner/configuration_spec.rb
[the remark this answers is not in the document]
JonRowe: Sorry I meant `class_double`.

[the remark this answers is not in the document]
botandrose: Oh, okay! But I can't seem to get that working either :rofl:

Can you say a bit more about how `class_double` could improve this spec? Sorry I'm so dense... I used to think I was pretty darn knowledgeable about RSpec!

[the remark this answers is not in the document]
JonRowe: `class_double("DatabaseCleaner::ActiveRecord::Truncation").as_stubbed_const` gives you a double which will only allow you to stub/expect methods that exist on the class, its a verifying double :)

botandrose:
@JonRowe @etagwerker Okay, I have removed the [WIP] from the PR title, because I think this is ready to go! There are a few issues that Jon brought up that remain unresolved, like exactly what to do with Gemfile.lock & dependabot, but I don't think those or anything else is blocking a v2.0.0.beta release! We can figure those out in the space between the beta and full release.

I have also created v2.0.0.beta PRs for the five adapters I think we should continue to support (the rest I think we can leave at 1.8):

* [database_cleaner-active_record](https://github.com/DatabaseCleaner/database_cleaner-active_record/pulls/1)
* [database_cleaner-sequel](https://github.com/DatabaseCleaner/database_cleaner-sequel/pulls/1)
* [database_cleaner-mongoid](https://github.com/DatabaseCleaner/database_cleaner-mongoid/pulls/1)
* [database_cleaner-redis](https://github.com/DatabaseCleaner/database_cleaner-redis/pulls/1)
* [database_cleaner-neo4j](https://github.com/DatabaseCleaner/database_cleaner-neo4j/pulls/1)

I know both of you are very busy, so any objections to me merging these PRs and cutting v2.0.0.beta releases in, let's say, three days?

spec/database_cleaner/configuration_spec.rb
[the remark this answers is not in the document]
botandrose: Ohhh, I see what you're getting at. I actually don't think that would be helpful here, because the `database_cleaner-active_record` gem isn't loaded at all! This suite is testing core in isolation from any of the adapters. That's why we need to stub these constants so that the strategy lookup code has something to find when we say `DatabaseCleaner[:active_record].strategy = :truncation`.

So, I don't think a verifying double would add much value, because the only methods being called on these constants is `const_get` on DC::AR to get DC::AR::Truncation, and `.new` on DC::AR::Truncation.

Now that I write this, it occurs to me that this situation could probably be improved by using an obviously fake set of constants, so that its clear that the implementation doesn't matter.

What do you think?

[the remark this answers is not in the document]
JonRowe: They behave like normal doubles if the class isn't loaded, they offer an extra level of protection when they are.

JonRowe:
I have no objections, but @etagwerker should have a say

spec/database_cleaner/configuration_spec.rb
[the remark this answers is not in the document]
botandrose: Wow, once again, I didn't know that! Very cool. I'm convinced. I'll give it a shot!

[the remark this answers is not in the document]
botandrose: Okay, pushed a new commit using `class_double` that I hope is making good use of it. Thanks for being a great teacher!

etagwerker:
@botandrose @JonRowe looking into it. Thanks for your hard work! ❤️

lib/database_cleaner/base.rb
```diff
@@ -48,8 +45,7 @@ def strategy
     attr_reader :orm
 
     def orm=(desired_orm)
-      @orm = (desired_orm || :autodetect).to_sym
-      @orm = @orm_autodetector.orm if @orm == :autodetect
+      @orm = desired_orm && desired_orm.to_sym
```
etagwerker: @botandrose I don't understand this line. Wouldn't this be the same?

```suggestion
      @orm = desired_orm.to_sym
```

spec/database_cleaner/configuration_spec.rb
```diff
@@ -9,131 +9,130 @@ def stub_cleaners(array)
...
+      expect(config.cleaners.values).to eq [cleaner]
     end
 
     it "should accept :data_mapper" do
-      cleaner = subject[:data_mapper]
+      cleaner = config[:data_mapper]
       expect(cleaner).to be_a(DatabaseCleaner::Base)
       expect(cleaner.orm).to eq :data_mapper
-      expect(subject.cleaners.values).to eq [cleaner]
+      expect(config.cleaners.values).to eq [cleaner]
```
etagwerker: @botandrose I thought you had decided to drop support for this. So maybe it's okay to remove this here?

spec/database_cleaner/configuration_spec.rb
```diff
@@ -9,131 +9,130 @@ def stub_cleaners(array)
...
     end
 
     it 'accepts :ohm' do
-      cleaner = subject[:ohm]
+      cleaner = config[:ohm]
       expect(cleaner).to be_a(DatabaseCleaner::Base)
       expect(cleaner.orm).to eq :ohm
-      expect(subject.cleaners.values).to eq [cleaner]
+      expect(config.cleaners.values).to eq [cleaner]
```
etagwerker: @botandrose Same comment as above.

lib/database_cleaner/base.rb
JonRowe: Nil raises an error if you do that.

.travis.yml
```diff
@@ -1,27 +1,13 @@
-dist: trusty
 language: ruby
 rvm:
-  - 1.9.3
-  - 2.0.0
-  - 2.1
-  - 2.2
```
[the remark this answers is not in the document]
botandrose: Right!?

database_cleaner-core.gemspec
```diff
@@ -0,0 +1,29 @@
+lib = File.expand_path("../lib", __FILE__)
+$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
+require "database_cleaner/version"
+
+Gem::Specification.new do |spec|
+  spec.name        = "database_cleaner-core"
+  spec.version     = DatabaseCleaner::VERSION
+  spec.authors     = ["Ben Mabey", "Ernesto Tagwerker"]
```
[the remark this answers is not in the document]
botandrose: Ah, that makes sense. Done! Thanks for bringing it up.

lib/database_cleaner/base.rb
botandrose: Hmm yeah. I've added a test case to the spec to ensure that passing nil doesn't cause it to crash.

spec/database_cleaner/configuration_spec.rb
botandrose: Yes, good catch! Pushed. Love seeing more red in a diff!

lib/database_cleaner/base.rb
botandrose: Although, now I'm thinking it should raise `NoORMDetected` like `DatabaseCleaner[nil]` does. What do you guys think?

botandrose: Probably should rename that exception, too, while we're at it, now that there's no more orm detection!

JonRowe: I mean I think its far to raise an `ArgumentError` here at least...

botandrose: Yeah, good call. That sounds like the right move to me. Replace NoORMDetected with ArgumentError, and raise it in both cases.

botandrose: Digging into this issue, I began to wonder if this is an opportunity to expose NullStrategy? For example:

```ruby
DatabaseCleaner[nil].clean # => Clean with NullStrategy
DatabaseCleaner.orm = nil # => Set to NullStrategy
DatabaseCleaner.strategy = nil # => Set to NullStrategy
```

This is getting a little out of scope of this PR, but what do y'all think?

botandrose:
@etagwerker @JonRowe I'm feeling pretty good. Any objection to merging this PR (and adapter PRs) into master and releasing a v2.0.0.beta series? Unresolved conversations here could be split off into their own issues to resolve before v2.0.0 final so we don't lose track of them. Yea or nay?

Gemfile.lock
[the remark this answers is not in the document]
JonRowe: Dependabot isn't up to that yet. It too often misses things because theres no security implication, or ignores things it can't update.

lib/database_cleaner/base.rb
JonRowe: I think the likely hood of someone doing that intentionally is lower than someone making a mistake and passing nil unintentionally, you'd still want an error or a warning somewhere even if you did that.

JonRowe:
No blocking objections

Gemfile.lock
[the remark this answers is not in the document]
botandrose: Ah, well, that's a bummer. Maybe some day. I'll remove the .lock files from git before the release.

botandrose:
Okay, I think everything has been resolved, now! Please let me know if not. Merging this into master.

JonRowe:
Great work 🙌

PR #625, opened 2020-02-10
Fix performance issue of DatabaseCleaner::Base#orm_module

### LINKED ISSUE #624: Performance degradation of `DatabaseCleaner.strategy=` in database_cleaner >= v1.8.0
With the recent refactoring of database_cleaner into multiple gems, the performance of `DatabaseCleaner.strategy=` took a big hit. As it is not unusual that the method will be called in a `before(:each)` hook, this can significantly slow down a test suite. (In our case it's about 30x slower)

A simple benchmark outlines the problem:

```ruby
# Database Cleaner 1.7.0
puts Benchmark.measure { 1000.times { DatabaseCleaner.strategy = :transaction } }
# => 0.006961   0.002179   0.009140 (  0.010197)

# Database Cleaner 1.8.0
puts Benchmark.measure { 1000.times { DatabaseCleaner.strategy = :transaction } }
# => 31.013500  26.446908  57.460408 ( 57.592854)
```

I haven't fully figured out the root cause, but I suspect it has something to do with how the strategies are required via RubyGems instead of just the file system.

benedikt:
This pull request fixes a performance degradation that ultimately caused DatabaseCleaner.strategy= to be very slow. Instead of requiring the correct gem for the given orm every time it is called, it first tries to get the constant and only requires the gem if the constant is missing.

The implementation is similar to how `DatabaseCleaner::Base#orm_strategy` works.

This pull request closes #624

### DIFF
--- a/lib/database_cleaner/base.rb
+++ b/lib/database_cleaner/base.rb
@@ -116,10 +116,12 @@ module DatabaseCleaner
 
     def orm_module
       return unless [:active_record, :data_mapper, :mongo, :mongoid, :mongo_mapper, :moped, :couch_potato, :sequel, :ohm, :redis, :neo4j].include?(orm)
+      orm_module_name = ORMAutodetector::ORMS[orm]
+      DatabaseCleaner.const_get(orm_module_name, false)
+    rescue NameError
       $LOAD_PATH.unshift File.expand_path("#{File.dirname(__FILE__)}/../../adapters/database_cleaner-#{orm}/lib")
       require "database_cleaner/#{orm}"
-      orm_module_name = ORMAutodetector::ORMS[orm]
-      DatabaseCleaner.const_get(orm_module_name)
+      retry
     end
 
     def orm_strategy(strategy)


benedikt:
@etagwerker Sure! It's not as fast as the initial version, but a lot faster than the current one.

```ruby
# Database Cleaner 1.7.0
puts Benchmark.measure { 1000.times { DatabaseCleaner.strategy = :transaction } }
# => 0.006961   0.002179   0.009140 (  0.010197)

# Database Cleaner 1.8.0
puts Benchmark.measure { 1000.times { DatabaseCleaner.strategy = :transaction } }
# => 31.013500  26.446908  57.460408 ( 57.592854)

# Database Cleaner (patched)
puts Benchmark.measure { 1000.times { DatabaseCleaner.strategy = :transaction } }
# => 0.109227   0.026194   0.135421 (  0.135726)
```

benedikt:
The failed specs are caused by the modules loaded via the version requirements in the gem specs. The `database_cleaner/#{orm}/version` files define the module, but not the default_strategy file. Once the gemspec requires `database_cleaner/#{orm}` the problem disappears.

benedikt:
I'm sorry, I give up. I can't get all the dependencies to compile (the mysql gem) or when I just remove that dependency, I run into a ton of weird load errors.

My suspicion is that the failing tests can be fixed by either making sure no `DatabaseCleaner::#{orm}` modules are created when only loading the `DatabaseCleaner::#{orm}::VERSION` or by explicitly loading all `DatabaseCleaner::#{orm}` classes in the spec_helper or something.

botandrose:
@benedikt Hello, and thanks for reporting this issue and supplying a PR! This autodetection code is all going to be deleted in v2.0.0, which should be released soon. But, it would be nice to not have a performance regression in the last 1.x release for those who might want to stay on 1.x for one reason or another! I'll take a look, and see if I can't get to the bottom of the test failures.

botandrose:
@benedikt Your hypothesis about the version file being loaded was indeed correct. I tried a few hacky ways to work around this issue, and wile I don't love any of them all that much, I feel okay with the following approach, given that this code is only going to live in the last bugfix releases of the 1.x line.

https://github.com/DatabaseCleaner/database_cleaner/compare/fix_orm_module_performance_regression?expand=1

It passes all the tests, and benchmarking on my machine seems to yield a similar speedup to your patch... i.e. about 10x slower than 1.7, but still much faster than 1.8. Would you mind verifying this branch yields the same performance improvement on your machine? If so, I'll merge it in and drop a v1.8 bugfix release.

benedikt:
@botandrose Hey! Thanks a lot for looking into this. I can confirm that your fix dramatically improves the performance. 

Looking forward to v2.0.0! :)

botandrose:
Okay, merged in via #627

botandrose:
https://rubygems.org/gems/database_cleaner/versions/1.8.3

Issue #628, opened 2020-02-28
Adding new adapters


TheSmartnik:
Hi!
Since database_cleaner is moving away from putting all adapters in `adapters` folder. I was wondering how to contribute new adapters? I guess would be a good idea to add a README entry for that.

etagwerker:
@TheSmartnik Definitely a good idea. Do you want to take a stab at it? 

The good news is that there are many examples out there now. For example: https://github.com/databaseCleaner/database_cleaner-active_record

I think the most important code will be in this folder: https://github.com/DatabaseCleaner/database_cleaner-active_record/tree/master/lib/database_cleaner/active_record

The section in the README should say a little bit about: 

- What strategies is your new adapter going to support? 
- What methods do you have to define in your new adapter?

Anything else, @botandrose?

TheSmartnik:
@etagwerker thank you for a quick reply! Sure, would love to do that. One thing is unclear, though. 
When adding a new adapter:

1. Do I create a folder `adapters` in database_cleaner repository and you create a separate gem and rept after the pr is merged.
2. Do I have to create my own repo and transfer it to you
3. Something else?

botandrose:
@TheSmartnik Yes, thank you for pointing this out! Now that adapters can be made and maintained without involving changes to database_cleaner core, we should absolutely document how this can be done in the README. Your perspective as an outsider trying to do this will be extremely helpful in creating this documentation. Thank you! I think @etagwerker's suggestion above is a really good place to start. Any bumps you hit or questions you have while doing this will be really valuable.

To answer your question above: in theory, there would be no actual need to involve database_cleaner core at all, in the form of PRs, etc. You can create your adapter repo, and maintain it separately, just as the split-off adapter gems are.

I suppose there's the question of having it be "blessed" or "adopted" into the DatabaseCleaner Github organization, but I think that question can be deferred until after the adapter is up and running and in use.

etagwerker:
@TheSmartnik What @botandrose said is correct: 

> 1. Do I create a folder adapters in database_cleaner repository and you create a separate gem and rept after the pr is merged.

You don't need to submit a PR to `database_cleaner` with the adapter code, you can create your own repository. Once it's up and running, you can submit a PR to `database_cleaner` so that we can list it here: https://github.com/databaseCleaner/database_cleaner/#list-of-adapters 

> 2. Do I have to create my own repo and transfer it to you

[...]

PR #629, opened 2020-03-05
Add shared example for database_cleaner adapter

TheSmartnik:
Following the discussion https://github.com/DatabaseCleaner/database_cleaner/issues/628
This pr adds shared examples for a generic adapter

### DIFF
--- a/lib/database_cleaner/spec/shared_examples.rb
+++ b/lib/database_cleaner/spec/shared_examples.rb
@@ -13,3 +13,12 @@ RSpec.shared_examples_for "a generic transaction strategy" do
   it { is_expected.to respond_to(:clean) }
   it { is_expected.to respond_to(:cleaning) }
 end
+
+RSpec.shared_examples_for "a database_cleaner adapter" do
+  it { expect(described_class).to respond_to(:available_strategies) }
+  it { expect(described_class).to respond_to(:default_strategy) }
+
+  it 'default_strategy should be part of available_strategies' do
+    expect(described_class.available_strategies).to include(described_class.default_strategy)
+  end
+end


botandrose:
@TheSmartnik This looks really good! Thank you!

Looking at the existing example blocks, I think we should take this opportunity to merge the three existing example blocks into a single `a database_cleaner strategy` shared example block, since there is no useful difference between them. Then, I'm wondering if its possible to have the new shared example block loop through the all the strategies defined in `available_strategies` and test to make sure they are valid strategies by testing each one against this new `a database_cleaner strategy` block.

This might be getting out of scope for this PR, though. We could save all this for another PR, if you just want to get this bit merged. Up to you.

TheSmartnik:
@botandrose thanks for a quick review!

> I think we should take this opportunity to merge the three existing example blocks into a single a database_cleaner strategy shared example block, since there is no useful difference between them. 

I like the idea, but sequel's adapter has a deletion strategy with a different API. It [doesn't have](https://github.com/DatabaseCleaner/database_cleaner-sequel/blob/master/lib/database_cleaner/sequel/deletion.rb#L36) `#start`/`#clean`/`#cleaning`, however it does [use](https://github.com/DatabaseCleaner/database_cleaner-sequel/blob/master/spec/database_cleaner/sequel/deletion_spec.rb#L8)  `a generic strategy` shared_example

botandrose:
@TheSmartnik Ah, thank you for taking such a deep look at this! I took a look at the Sequel deletion adapter, and it does actually have those methods, even though it doesn't really look like it. The inheritance and inclusion heirarchy is kind of a mess, and makes it hard to see where things are coming from! It inherits `#start` from its immediate ancestor, the `DatabaseCleaner::Sequel::Truncation` strategy, and it has `#cleaning` included from `DatabaseCleaner::Generic::Base` which is included in `DatabaseCleaner::Sequel::Base`, which is also included in its immediate ancestor. What a confusing mess!

In any event, the test you pointed out should definitely be updated to use our new shared examples, after this is merged!

botandrose:
@TheSmartnik Looks really great! Thank you for your work here.

PR #636, opened 2020-04-05
Reenable E2E cucumber tests

botandrose:
* Strip down cucumber scenarios to just use the two most popular adapter gems
* Remove tests for adapters included in v1.8 database_cleaner gem
* Reenable cucumber scenarios in default rake task, and therefore in Travis
* Remove last remnant of removed v1.x ORM autodetection, exposed by E2E tests.


botandrose:
I'm having major issues with Travis, so I'm going to close this until I figure those out.

PR #639, opened 2020-04-06
Add changelog to gemspec

rafbm:
For a nice “Changelog” link in the https://rubygems.org/gems/database_cleaner sidebar.


database_cleaner.gemspec
```diff
@@ -11,6 +11,8 @@ Gem::Specification.new do |spec|
   spec.homepage    = "https://github.com/DatabaseCleaner/database_cleaner"
   spec.license     = "MIT"
 
+  s.metadata["changelog_uri"] = "https://github.com/DatabaseCleaner/database_cleaner/blob/master/History.rdoc"
```
botandrose, line 14: Great idea, but the `s` variable needs to be `spec`. Fix that and I think this is good to merge!

botandrose:
I'm guessing this won't show up until the next gem release?

rafbm:
I guess so.

PR #641, opened 2020-04-07
Feature/codecov coverage

### LINKED ISSUE #638: Test coverage tracking?
Hey @etagwerker,

I would like to get some test coverage tracking going, probably either with Coveralls or Code Climate, but there's the question of permissions.

I could set up Coveralls myself, but then I think my @botandrose account would be the only account with permissions, which is not ideal.

Code Climate is already set up, but I don't have the permissions to add test coverage to it.

Do you know of a way to give the DatabaseCleaner organization (and therefore its members) access to one or the other? Or some other solution?

etagwerker:
Hey @botandrose, 

This PR adds `simplecov` + `codecov` to this project. It fixes #638. 

Please check it out.

Thanks!


botandrose:
@etagwerker This looks great! Thank you for taking the time to put this together!

PR #643, opened 2020-04-07
[BREAKING CHANGE] Remove all #orm= setter methods

botandrose:
With the decomposition of this gem into individual ORM adapter gems, users of this library are now explicitly opting in to which ORMs they are using by virtue of which adapter gems they require. That -- plus the fact that ORM autodetection has been removed -- makes the `#orm=` setter methods at the various levels of the DC object graph rather useless at best, and problematic at worst, IMO. Lastly, note that there's no diff in the README, because this method was never advertised in it (AFAICT with git log -p), so that helps things a bit.

Still, it is a breaking change, and while we are gearing up for a breaking release with v2.0.0, this method has not been deprecated in the v1.x series.

Assuming that we do indeed want to follow through with this API removal, I can think of three paths forward:
1. Do nothing. Simply release v2.0.0 without the #orm= methods.
2. Push a deprecation warning into a v1.8.x release.
3. Hold off on a new v1.8.x release, but push a deprecation warning into a new branch for a v1.9 release that will contain this deprecation, and any other we might discover we want to do before v2.0.

@etagwerker @JonRowe What do you guys think about all this? Any reason we might not want to remove this? Other considerations or ideas?

### DIFF
--- a/lib/database_cleaner/cleaner.rb
+++ b/lib/database_cleaner/cleaner.rb
@@ -13,11 +13,13 @@ module DatabaseCleaner
       [orm, db] <=> [other.orm, other.db]
     end
 
-    def initialize(orm = :null, opts = {})
-      self.orm = orm
+    def initialize(orm, opts = {})
+      @orm = orm
       self.db = opts[:connection] || opts[:model] if opts.has_key?(:connection) || opts.has_key?(:model)
       Safeguard.new.run
     end
+
+    attr_reader :orm
 
     def db=(desired_db)
       @db = self.strategy_db = desired_db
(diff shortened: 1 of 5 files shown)


JonRowe:
The approach we took with RSpec which had a large amount of breaking changes was to release a `2.99.0` series, which was intended to be a last "deprecation" version. This meant everyone updating would get a final version before modifying their Gemfile to run on RSpec 3. We also made the commitment that a passing, deprecation free build on 2.99 would run on 3.

So I'd pick option 3. It worked very well.

etagwerker:
@botandrose I like the idea described in option 3. I also like the way that RSpec managed the 3.0 release. I wonder if we could do a 1.99 release with all these deprecation warnings. That way we get to a point where: 

- 1.99 has all the deprecation warnings pre-2.0
- 2.0 removes all deprecation warnings

botandrose:
Thanks for the input, guys! I remember the RSpec 2.99 -> 3.0 dance, and I also remember how dang smooth it was, so I'm happy to steal that idea for this library's 2.0 journey. I'll make a v1.99 branch today, and put this deprecation warning in it.

PR #644, opened 2020-04-11
New API for ORM Adapter gems

botandrose:
## Motivation
Since v2.0.0 will be the first time we're actually allowing external adapter gems to be created (the list of recognized adapters is hard-coded in the v1 line), this API addition is technically only a backwards-compatible semver-minor change. However, we will be committing to supporting whatever ORM Adapter API we release in v2.0.0 until at least v3.0.0, so I think its really important that we get this right before the first v2 release. Therefore I'd really appreciate your feedback!

To test out this new API, I have also opened two companion PRs for the two most popular ORM adapter gems: ActiveRecord and Redis. These also happen to be particularly good candidates for evaluating this proposal, because they are the most complex adapter and the simplest adapter, respectively.

https://github.com/DatabaseCleaner/database_cleaner-active_record/pull/41
https://github.com/DatabaseCleaner/database_cleaner-redis/pull/8

I'll start by describing the current state of affairs in the Status Quo section, and then afterwards, I'll describe the changes proposed in these PRs in the Proposal section.

### ORM Adapter
The ORM Adapter namespace module must define two module functions, e.g.:

```ruby
# lib/database_cleaner/active_record.rb
require "database_cleaner/active_record/version"
require "database_cleaner/core"
require "database_cleaner/active_record/transaction"
require "database_cleaner/active_record/truncation"
require "database_cleaner/active_record/deletion"

module DatabaseCleaner::ActiveRecord
  def self.available_strategies
    %i[transaction truncation deletion]
  end

  def self.default_strategy
    :transaction
  end
end

# explicit registry, now that ORM autodetection is out
DatabaseCleaner[:active_record].strategy = :transaction
```

### Strategies
Strategies get their functionality in bits and pieces by including various modules (which themselves include other modules). In core, there are three different modules in three different files under `DatabaseCleaner::Generic`:
1. `DatabaseCleaner::Generic::Base`
2. `DatabaseCleaner::Generic::Transaction`
3. `DatabaseCleaner::Generic::Truncation`

Understanding which to include where is surprisingly tricky, leading to the code in one single strategy being spread across four files and two gems! For example, here is the v1.8 module inclusion heirarchy of the truncation strategies from the ActiveRecord and Redis adapters.

```
class DatabaseCleaner::ActiveRecord::Truncation
├--> module DatabaseCleaner::ActiveRecord::Base
|    └--> module DatabaseCleaner::Generic::Base
└--> module DatabaseCleaner::Generic::Truncation
```

```
class DatabaseCleaner::Redis::Truncation
├--> module DatabaseCleaner::Redis::Base
|    └--> module DatabaseCleaner::Generic::Base
└--> module DatabaseCleaner::Generic::Truncation
```

## Proposal
Elucidated in this PR and companion PRs:
https://github.com/DatabaseCleaner/database_cleaner-active_record/pull/41
https://github.com/DatabaseCleaner/database_cleaner-redis/pull/8

### ORM Adapter
Both module functions are no longer necessary! The default strategy is already set explicitly when the adapter registers itself, and the available strategies are now determined through introspection of the adapter module namespace. One downside to this is that internal adapter constants are now falsely interpreted as strategies, but those can be privatized to avoid this. See DC::AR PR for an example of this.

Honestly, I'm not 100% on this exact strategy, and I wouldn't be surprised if this particular bit saw some iteration before final release. For example, we could alternatively do the detection by checking for descendence from DC::Strategy, etc. Despite this uncertainty, I think the goal of narrowing the Adapter API to its minimum is really worth shooting for, even perhaps to the point of erring a bit too far in v2.0! Because we can always change our minds and add something back in v2.1 or v2.2, etc, but we cannot take anything away until at least v3.0, which I hope is a long way off.

```ruby
# lib/database_cleaner/active_record.rb
require "database_cleaner/active_record/version"
require "database_cleaner/core"
require "database_cleaner/active_record/transaction"
require "database_cleaner/active_record/truncation"
require "database_cleaner/active_record/deletion"

DatabaseCleaner[:active_record].strategy = :transaction
```

### Strategies
All three modules in `DatabaseCleaner::Generic` are superseded by a new single `DatabaseCleaner::Strategy` abstract class, and the concrete strategies simply inherit from that, now. The examples above become: 

```
class DatabaseCleaner::ActiveRecord::Truncation
└--> class DatabaseCleaner::ActiveRecord::Base
     └--> class DatabaseCleaner::Strategy
```

```
class DatabaseCleaner::Redis::Truncation
└--> class DatabaseCleaner::Strategy
```

This can be done because the differences between the original three generic modules were largely superficial, with the exception of the `DatabaseCleaner::Generic::Truncation` module doing some work to enable the `:only` and `:except` options, as those are likely to be useful in every truncation strategy. To address this I have refactored this option handling into a private `DatabaseCleaner::Strategy#tables_to_clean` method that is included in every strategy. If those two options are useful to a strategy, then this method can be used internally to provide them. If not, it can simply be ignored! You can see this method being used in both of the truncation strategies in the two adapter PRs.

## Conclusion
I think this PR is an substantial improvement over the status quo, particularly with regards to simplifying the developer experience of creating an ORM adapter gem, and minimizing the surface area of the API. However, there are certainly lots of alternative designs to what I have proposed here, both in the large, and in the details, and I want to get this right in v2.0.0, so I'd love to receive your feedback!


botandrose:
I've pushed a few more commits which make two minor changes to make the ORM Adapter API even simpler, still.

* Strategies now _must_ be a descendant of `DatabaseCleaner::Strategy`. As mentioned above, this cuts down on false positives in the detection of available strategies, which mostly means that adapter authors probably won't have to know how it works, or need to muck with `private_constant`. Simpler!

* Remove the private `#tables_to_clean ` method. After looking more closely at this method, it seemed to me that it wasn't at all pulling its weight. It only does two things:
  1. Throws an error if both `only` and `except` were passed, explaining that it doesn't really make sense. IMO, this is of dubious value. I actually disagree with the assertion that that situation is nonsensical; saying "clean these, but not those" makes perfect sense to me! So I've simply removed it. If someone manages to come up with a scenario that demonstrates this kind of check is actually needed, it can be put into place at the adapter level, instead.
  2. Performs a single array subtraction.
IMO, neither of those things are worth the cost of adding that method to the adapter API, nor the cost of the indirection that using that method would entail. Removing the check, and inlining the subtraction it into both adapter PRs shows an immediate improvement in grokkability, and surprisingly without an increase in complexity. `Strategy`, now, is dead simple. It seems like an obvious move now, in hindsight.

With these two changes, I'm personally very happy with the state of the ORM Adapter API, and think its ready to merge. Since there have been no issues raised here, I'm planning on merging this tomorrow, unless someone chimes in with a concern. The next (and hopefully final) beta release will be forthcoming soon after that!

PR #645, opened 2020-05-04
Fix :mongo strategy

mtsmfm:
I send this patch to the maintenance branch because it's broken even on stable release (especially the cause 1 is introduced in 1.8.0).

https://github.com/DatabaseCleaner/database_cleaner/commit/43d6482fb2b7194c554d1f32a0d50b457568f3ac

Please feel free to ask me if I should change the base branch or send the same patch to master branch (and adapter repo) as well.

----

Currently, the following code doesn't work:

```ruby
DatabaseCleaner[:mongo].strategy = :truncation

RSpec.configure do |config|
  config.around do |example|
    DatabaseCleaner.cleaning do
      example.run
    end
  end
end
```

The causes are:

1. `DatabaseCleaner::ORMAutodetector` doesn't consider `:mongo` option
2. `DatabaseCleaner::Mongo::Truncation` doesn't respond to `#cleaning`

### DIFF
--- a/adapters/database_cleaner-mongo/lib/database_cleaner/mongo/truncation.rb
+++ b/adapters/database_cleaner-mongo/lib/database_cleaner/mongo/truncation.rb
@@ -1,9 +1,11 @@
 require 'database_cleaner/mongo/base'
+require 'database_cleaner/generic/base'
 require 'database_cleaner/generic/truncation'
 require 'database_cleaner/mongo/truncation_mixin'
 module DatabaseCleaner
   module Mongo
     class Truncation
+      include ::DatabaseCleaner::Generic::Base
       include ::DatabaseCleaner::Generic::Truncation
       include TruncationMixin
       include Base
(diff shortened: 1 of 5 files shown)


botandrose:
Wow, great catch, @mtsmfm! Thank you so much for the excellent detective work, and for the flawless PR.

I don't _believe_ any of this will need to be ported to master on database_cleaner or database_cleaner-mongo, since autodetection was completely removed from the former, and the new adapter api should have provided the missing `#cleaning` method in the latter. But feel free to double-check... as this regression has demonstrated, I can make mistakes! XD

I'll go ahead cut new bugfix releases of both gems, presently, to get this fix out the door.

Thanks again!

botandrose:
https://rubygems.org/gems/database_cleaner/versions/1.8.5
https://rubygems.org/gems/database_cleaner-mongo/versions/1.8.1

PR #646, opened 2020-05-07
Fix :mongo adapter to clean up newly created collection

mtsmfm:
Currently, :mongo adapter only cleanups collections which exist at the first call of `#clean` because it caches collections.
But both :mongo_mapper and :mongoid adapter doesn't cache.
I confirmed it's slightly faster than non cached one but it costs only 5%.
I think we can accept removing this cache to cleanup all collections.

For some reason, this cache behavior considers mongoid but it doesn't depend on this adapter
so we can just remove these codes.

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'

  gem 'mongo'
  gem 'benchmark-ips', require: 'benchmark/ips'
  gem 'database_cleaner-mongo', '1.8.1'
end

Mongo::Logger.level = Logger::WARN
client = Mongo::Client.new('mongodb://mongo')

strategy1 = DatabaseCleaner::Mongo::Truncation.new.tap {|s| s.db = client.database }
strategy2 = DatabaseCleaner::Mongo::Truncation.new.tap {|s| s.db = client.database }.extend(Module.new do
  def collections
    database.collections.select { |c| c.name !~ /^system\./ }
  end
end)

collection_names = 1000.times.map {|i| "test_#{i}" }
(collection_names - client.collections.map(&:name)).each do |name|
  client[name].create
end

Benchmark.ips do |x|
  x.time = 10

  x.report("with cache") do
    strategy1.clean
  end

  x.report("without cache") do
    strategy2.clean
  end

  x.compare!
end

__END__

$ ruby bench.rb
Warming up --------------------------------------
          with cache     1.000  i/100ms
       without cache     1.000  i/100ms
Calculating -------------------------------------
          with cache      0.835  (± 0.0%) i/s -      9.000  in  10.803207s
       without cache      0.798  (± 0.0%) i/s -      8.000  in  10.034118s

Comparison:
          with cache:        0.8 i/s
       without cache:        0.8 i/s - 1.05x  (± 0.00) slower

```

### DIFF
--- a/adapters/database_cleaner-mongo/lib/database_cleaner/mongo/truncation.rb
+++ b/adapters/database_cleaner-mongo/lib/database_cleaner/mongo/truncation.rb
@@ -15,49 +15,8 @@ module DatabaseCleaner
         db
       end
 
-      def collections_cache
-        @@collections_cache ||= {}
-      end
-
-      def mongoid_collection_names
-        @@mongoid_collection_names ||= Hash.new{|h,k| h[k]=[]}.tap do |names|
-          ObjectSpace.each_object(Class) do |klass|
-            (names[klass.db.name] << klass.collection_name) if valid_collection_name?(klass)
-          end
-        end
-      end
-
-      def not_caching(db_name, list)
-        @@not_caching ||= {}
-
-        unless @@not_caching.has_key?(db_name)
-          @@not_caching[db_name] = true
-
-          puts "Not caching collection names for db #{db_name}. Missing these from models: #{list}"
-        end
-      end
-
       def collections
-        return collections_cache[database.name] if collections_cache.has_key?(database.name)
-        db_collections = database.collections.select { |c| c.name !~ /^system\./ }
-
-        missing_collections = mongoid_collection_names[database.name] - db_collections.map(&:name)
-
-        if missing_collections.empty?
-          collections_cache[database.name] = db_collections
-        else
-          not_caching(database.name, missing_collections)
-        end
-
-        db_collections
-      end
-
-      private
-
-      def valid_collection_name?(klass)
-        klass.ancestors.map(&:to_s).include?('Mongoid::Document') &&
-        !klass.embedded &&
-        !klass.collection_name.empty?
+        database.collections.select { |c| c.name !~ /^system\./ }
       end
     end
   end
--- a/spec/database_cleaner/mongo/mongo_examples.rb
+++ b/spec/database_cleaner/mongo/mongo_examples.rb
@@ -23,4 +23,6 @@ module MongoTest
   end
   class Gadget < Base
   end
+  class Gizmo < Base
+  end
 end
--- a/spec/database_cleaner/mongo/truncation_spec.rb
+++ b/spec/database_cleaner/mongo/truncation_spec.rb
@@ -28,6 +28,20 @@ RSpec.describe DatabaseCleaner::Mongo::Truncation do
       end
     end
 
+    context "when new collection is created after clean" do
+      before do
+        subject.clean
+
+        MongoTest::Gizmo.new(name: 'some gizmo').save!
+      end
+
+      it "truncates new collection" do
+        expect { subject.clean }.to change {
+          MongoTest::Gizmo.count
+        }.from(1).to(0)
+      end
+    end
+
     context "when collections are provided to the :only option" do
       subject { described_class.new(only: ['MongoTest::Widget']) }
 


botandrose:
@mtsmfm Hiya, thanks for another excellent and well-thought-through PR! I agree that this is a surprising behavior, and think we should do something about it.

I'm a little uncertain about how to consider this change, though, in terms of SemVer. I can think of three possibilities:

1. Bugfix. We could consider removing the caching behavior to be fixing a bug. However, I can imagine existing test suites implicitly depending on this caching behavior, and breaking once its "fixed," so at the very least this would be a breaking bugfix.
2. Minor. We could reframe this in terms of adding a new `:cache_collections` strategy option, similar to how the SQL-based adapters' `:cache_tables` option works. It would be `true` by default, for backwards compatibility, and we could switch it to `false` by default in v2.0.
3. Major. We just merge this into the upcoming v2.0 release, and consider it a breaking change. If someone misses the old caching behavior, they can add it back in as that `:cache_collections` option for a later v2.x release.

Considering these options, I am wary of potential breakage from option 1, so I'm personally leaning towards either option 2 or 3.  I'm thinking the choice between those two ultimately comes down to the value of caching the collections. Your benchmarking above seems to indicate that the speedup gained by caching is relatively negligible, which makes it seem unlikely that anyone would want to enable the collection caching options. This thinking leads me to conclude that option 3 is probably the best way forward.

What do you think?

mtsmfm:
I think option 1 is preferable.

1. Randomizing test execution order is one of the often-used techniques. But in such a case, this caching behavior can make tests flaky. Let's say we have the following test:

```ruby
describe User do
  it do
  end
end

describe Post do
end
```
I think this fix makes tests stable.

mtsmfm:
@botandrose What do you think?
If you disagree with option 1, I'll create another PR to follow your decision.

botandrose:
@mtsmfm Sorry for the late reply here. Thank you for your time and attention!

It sounds like this collection name caching issue is even worse than I first imagined. It sounds like it is likely affecting everyone who is using the mongo adapter for database_cleaner! Is this true, do you think? Did this mongo adapter ever really work well at all? When did this caching behavior start?

mtsmfm:
> Is this true, do you think?

I think so

>  Did this mongo adapter ever really work well at all?

I'm not sure, I recently started using mongo adapter.
It seems reason 2 shows almost no user uses this adapter though.

https://github.com/DatabaseCleaner/database_cleaner/pull/646#issuecomment-626116003

> When did this caching behavior get introduced to the codebase?

It seems from v1.4.0.
https://github.com/DatabaseCleaner/database_cleaner/pull/110

At that time, mongo and mongoid adapters shared some codes and I guess #110 didn't consider pure mongo adapter.
https://github.com/DatabaseCleaner/database_cleaner/blob/v1.4.0/lib/database_cleaner/mongoid/truncation.rb#L16

----

JFYI:

mongo adapter was introduced in v0.9.0 and there's no cache between from v0.9.0 to v1.4.0

https://github.com/DatabaseCleaner/database_cleaner/pull/138

botandrose:
@mtsmfm Wow, great detective work. Thank you! It really helps to have all this historical context.

So, to summarize my understanding: It seems likely that the :mongo adapter is basically unusable for most test suites, and has been this way ever since v1.4.0, eight years ago. This PR would fix it and make it usable again, but if there is anyone actually using this adapter, this bugfix will probably require them to make changes to their app or test suite. Does this all sound right to you?

I try to adhere to Semver as much as possible, so the fact that this bugfix would require at least some users to change their code, forces a major version bump. I know I would be surprised and 
upset if updating to a patch version of a gem broke something in my app! Luckily, we are very very close to releasing v2.0 of database_cleaner and all its adapter gems, so I think we should just leave the v1 mongo adapter as it has been for eight years, and make this big change in v2.0. We're also planning on making a v1.99 release with some more deprecation warnings, so we should make sure  to put a warning about this behavior change there, too, for the users that may still be using the mongo adapter. Maybe some are still stuck on v1.3!

So, yeah, please move this PR over to https://github.com/DatabaseCleaner/database_cleaner-mongo, and I'll merge it!

mtsmfm:
Okay, I created PR to mongo repo

https://github.com/DatabaseCleaner/database_cleaner-mongo/pull/6

mtsmfm:
@botandrose Sorry for bothering you but I'm waiting for 2.0 release while almost one year.
Do you have a plan to release 2.0 in the near future or can I ask you to include this PR as a bugfix?

botandrose:
hey @mtsmfm Sorry for long wait here. I think we are ready to release v2.0 of all the gems! Unfortunately, I switched phones some time last year and have been locked out of my Rubygems account ever since, due to 2FA. Every so often I try to regain access, and fail. I'll give it another go today.

botandrose:
@mtsmfm v2.0 is out!

PR #650, opened 2020-05-18
[BREAKING] Rename :connection configuration option to :db for consistency

### LINKED ISSUE #649: `:connection` vs `#db` confusion. Let's pick one!
@etagwerker These are two ways of saying the same thing, but they look pretty different:

```ruby
DatabaseCleaner[:active_record, connection: :logs]
DatabaseCleaner[:active_record].db = :logs
```

Ignoring, for the moment, that perhaps we could limit this to one (and only one) way, I think we should at least align the naming for v2.0. Options I can think of:

```ruby
DatabaseCleaner[:active_record, db: :logs]
DatabaseCleaner[:active_record].db = :logs
```

```ruby
DatabaseCleaner[:active_record, connection: :logs]
DatabaseCleaner[:active_record].connection = :logs
```

I'm leaning towards `db`, because its shorter, and (at least to me) has a broader connotation, where `connection` makes me think we could only pass a connection object or connection string. This is pretty subjective, for sure, though.

I think whichever one we go with, we could do it in a backwards-compatible way with a deprecation warning.

What do you think?

botandrose:
Renames the `:model` option, too, which was simply an alias, and shouldn't be in core, anyways, since it was ActiveRecord-specific. Will deprecate that one, too, in v1.99.

Closes #649

With this PR, and the truncation -> deletion renaming PRs in a few of the adapters, I think v2.0 is feature-complete!! I'm hoping to release both v1.99.0.beta and v2.0.0.beta versions of everything this weekend.

Its happening!!

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -294,7 +294,7 @@ For more examples see the section ["Why?"](#why).
 
 Sometimes you need to use multiple ORMs in your application.
 
-You can use DatabaseCleaner to clean multiple ORMs, and multiple connections for those ORMs.
+You can use DatabaseCleaner to clean multiple ORMs, and multiple databases for those ORMs.
 
 ```ruby
 require 'database_cleaner/active_record'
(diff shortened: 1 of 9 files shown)


botandrose:
@etagwerker yes, I agree 100%. I'm imagining v1.99 to be that interim release that supports both :connection (and :model) and :db, with a deprecation warning telling folks to just use :db. I'm going to branch v1.99 off of the v1-maintenance branch, so master is already v2.0, so I think this should be good to merge, even before v1.99 is released. Sound good to you?

PR #658, opened 2020-06-18
Add more descriptive "allowlist" terminology

### LINKED ISSUE #626: Rename whitelist to allowlist
I think it would be a friendlier term for the safeguard configuration: https://github.com/DatabaseCleaner/database_cleaner#safeguards

bpo:
Hi there, thanks for maintaining DatabaseCleaner!

This runs `s/url_whitelist/url_allowlist/` and adds backwards-compatible
aliases for the terminology which can be easily deprecated or dropped
in the future, should that be desired.

This also replaces the error `NotWhitelistedUrl` with the more descriptive
`UrlNotAllowed`. I believe merging will resolve #626.

### DIFF
--- a/History.rdoc
+++ b/History.rdoc
@@ -1,3 +1,11 @@
+== Development (unreleased)
+
+=== Changes
+  * Rename `url_whitelist` to `url_allowlist`
+
+=== Breaking changes
+  * Failed checks against the allowlist now raise `UrlNotAllowed` rather than `NotWhitelistedUrl`
+
 == 2.0.0.beta2 2020-05-30
 
 === Features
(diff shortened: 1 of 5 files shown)


bpo:
I think the build failure here is being caused by a new version of Cucumber and is not directly related to my proposed change.

botandrose:
@bpo Looks good to me!

Technically, this is not 100% backwards compatible, in the case that someone is doing something like `rescue NotWhitelistedUrl`, but I think that's highly unlikely, and our next release is a major release anyways. So I think its just fine as-is!

Two requests before merging:
1. I fixed the CI issue with cucumber on master, so please rebase and force-push so that we can see the magical green checkbox on this PR!
2. Can you add a changelog entry to History.rdoc?

Thanks for the PR!

bpo:
I think this is ready to go now. Thanks!

etagwerker:
@botandrose @bpo wonderful, thank you! ❤️

PR #660, opened 2020-06-30
Support regular expressions in allowlist

bpo:
It's common for database connection parameters to be specified as part of the URL (e.g. `DATABASE_URL=postgres://foo.bar?pool=16`). I'd like to be able to manipulate those connection parameters in my test rig without needing to alter the allowlist.

This patch supports this by letting allowlists optionally be specified as regular expressions.

### DIFF
--- a/History.rdoc
+++ b/History.rdoc
@@ -2,6 +2,7 @@
 
 === Changes
   * Rename `url_whitelist` to `url_allowlist`
+  * Allowlist now supports regular expressions
 
 === Breaking changes
   * Failed checks against the allowlist now raise `UrlNotAllowed` rather than `NotWhitelistedUrl`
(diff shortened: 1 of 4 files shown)


botandrose:
@bpo Thanks for the PR! I really like this change, and with the switch to `===`, I believe that one could even put a `proc` (or some other callable) in the list, as well, which is pretty cool.

One last thing before I sign off on it: would you please add a changelog entry?

@etagwerker This looks good to me. What do you think?

bpo:
:+1: added changelog, and another example using a `proc` to make sure that functionality sticks around.

etagwerker (approved):
@bpo @botandrose This looks good to me too. Thank you! ❤️

Issue #671, opened 2021-02-01
2.0 Regression: allow_remote_database_url and url_allowlist not working anymore


tmaier:
Dependabot just upgraded `database_cleaner-active_record` to 2.0 and now all tests fail.
I used to have `DatabaseCleaner.allow_remote_database_url = true` in my code.

<details>
<summary>Open to see error message and trace log</summary>

```text
Failure/Error: require File.expand_path('../config/environment', __dir__)

DatabaseCleaner::Safeguard::Error::RemoteDatabaseUrl:
  ENV['DATABASE_URL'] is set to a remote URL. Please refer to https://github.com/DatabaseCleaner/database_cleaner#safeguards
# /usr/local/bundle/gems/database_cleaner-core-2.0.0/lib/database_cleaner/safeguard.rb:45:in `run'
# /usr/local/bundle/gems/database_cleaner-core-2.0.0/lib/database_cleaner/safeguard.rb:104:in `block in run'
# /usr/local/bundle/gems/database_cleaner-core-2.0.0/lib/database_cleaner/safeguard.rb:104:in `each'
# /usr/local/bundle/gems/database_cleaner-core-2.0.0/lib/database_cleaner/safeguard.rb:104:in `run'
# /usr/local/bundle/gems/database_cleaner-core-2.0.0/lib/database_cleaner/cleaner.rb:31:in `initialize'
# /usr/local/bundle/gems/database_cleaner-core-2.0.0/lib/database_cleaner/cleaners.rb:41:in `new'
# /usr/local/bundle/gems/database_cleaner-core-2.0.0/lib/database_cleaner/cleaners.rb:41:in `add_cleaner'
# /usr/local/bundle/gems/database_cleaner-core-2.0.0/lib/database_cleaner/cleaners.rb:12:in `block in []'
# /usr/local/bundle/gems/database_cleaner-core-2.0.0/lib/database_cleaner/cleaners.rb:12:in `fetch'
# /usr/local/bundle/gems/database_cleaner-core-2.0.0/lib/database_cleaner/cleaners.rb:12:in `[]'
# /usr/local/bundle/gems/database_cleaner-active_record-2.0.0/lib/database_cleaner/active_record.rb:7:in `<main>'
# /usr/local/bundle/gems/bootsnap-1.6.0/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:23:in `require'
[...]

PR #673, opened 2021-02-03
Delay running safeguards until right before cleaning begins

botandrose:
This allows for time to configure safeguards. Addresses #671


etagwerker (approved):
@botandrose This looks good to me. Thanks! 💯

PR #675, opened 2021-02-11
Fixed readme

vitalinfo:
if use it as it described in Readme get an exception like
```
ArgumentError:
  wrong number of arguments (given 2, expected 1)
```
on Ruby 3


botandrose:
Thank you!

PR #681, opened 2021-03-16
Fixed strategy class lookup

### LINKED ISSUE #680: Poor strategy class lookup
You should use `strategy.to_s.classify` instead of `strategy.to_s.capitalize` in `DatabaseCleaner::Cleaner#orm_strategy`.
https://github.com/DatabaseCleaner/database_cleaner/blob/a34c4c3d94d7397878e3510987093f17603f81cf/lib/database_cleaner/cleaner.rb#L89

softpro:
Strategy name should be `camelize`d, not `capitalize`d.

Fixes #680.

### DIFF
--- a/lib/database_cleaner/cleaner.rb
+++ b/lib/database_cleaner/cleaner.rb
@@ -86,7 +86,7 @@ module DatabaseCleaner
     end
 
     def orm_strategy(strategy)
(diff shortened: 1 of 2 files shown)


botandrose:
Great work! Thank you!

PR #682, opened 2021-04-01
Fix docs about usage of only and except options

oboxodo:
The existing docs lead to non-working code at least with Ruby 2.7.

We need to either make the Array or the Hash explicit in order for it to work:

```
irb(main):026:0> DatabaseCleaner[:active_record].strategy = :truncation, only: ["users"]
Traceback (most recent call last):
SyntaxError ((irb):26: syntax error, unexpected tLABEL)
...].strategy = :truncation, only: ["users"]
...                          ^~~~~
irb(main):027:0> DatabaseCleaner[:active_record].strategy = [:truncation, only: ["users"]]
=> [:truncation, {:only=>["users"]}]
irb(main):028:0> DatabaseCleaner[:active_record].strategy = :truncation, {only: ["users"]}
=> [:truncation, {:only=>["users"]}]
```

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -75,11 +75,11 @@ DatabaseCleaner.clean
 With the `:truncation` strategy you can also pass in options, for example:
 
 ```ruby
-DatabaseCleaner.strategy = :truncation, only: %w[widgets dogs some_other_table]
+DatabaseCleaner.strategy = [:truncation, only: %w[widgets dogs some_other_table]]
 ```
 
 ```ruby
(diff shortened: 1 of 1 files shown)


viktorianer:
For me, it worked this way to:

```
DatabaseCleaner[:active_record].strategy = :transaction
DatabaseCleaner.clean_with(:deletion, except: %w(templates))
```

Either one way or another, it is incorrectly documented and needs to be corrected!

PR #697, opened 2022-12-27
Add Ruby 3.2 to the CI matrix

petergoldstein:
In addition to adding Ruby 3.2 to the CI matrix, I updated the version of the checkout action.

To get the specs to green I had to make a few additional changes:

1. Add explicit hashes in the `with` clauses to remove ambiguity with keyword arguments
2. Update the safeguard logic to treat an empty string for `host` as local.

This runs green on my fork.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,7 @@ jobs:
     strategy:
       matrix:
         ruby-version:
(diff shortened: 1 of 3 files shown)


etagwerker (approved):
@petergoldstein Looks good, thank you! 👍🏼

PR #703, opened 2023-08-11
Generalize ProductionEnv safeguard error to NonTestEnv.

cherbst-2112:
Expanding on the great idea in https://github.com/DatabaseCleaner/database_cleaner/pull/521, I was surprised to find database cleaner will happily delete data in development database if RAILS_ENV is set to development. This was a problem for us when working with a remote development database through an SSH tunnel. While the existing allow_remote_database_url config could have prevented this, the SSH tunnel being on localhost prevents the existing safeguard from working

So I thought checking the RAILS_ENV would be a good way to prevent this. While it's great that we already prevent doing this in production, there's really no reason to allow deleting data anywhere but test env so generalized this to error on any non-test environment.

I thought about adding a parallel implementation and leaving allow_production in place for existing users but anyone running with that configuration should immediately get an error with this and can then migrate to allow_non_test_env.

### DIFF
--- a/History.rdoc
+++ b/History.rdoc
@@ -1,5 +1,8 @@
 == Development (unreleased)
 
+=== Changes
+  * Safeguard against running in non-test environment
+
 == 2.0.2 2023-03-10
 
 == Bugfixes
--- a/README.markdown
+++ b/README.markdown
@@ -318,7 +318,7 @@ After copying and pasting code to do this several times I decided to package it 
 
 DatabaseCleaner comes with safeguards against:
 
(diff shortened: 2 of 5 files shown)


etagwerker:
@cherbst-2112 Thanks for taking the time to submit this PR. 

I thought about it and I don't think this is a good idea. 

There is already a mechanism to allow a list of database urls: `DatabaseCleaner.url_allowlist`

If you're afraid of deleting your development database, you can use that feature.

cherbst-2112:
OK sure, the host alone won't be enough since the danger lies in using a local SSH proxy (and I think this is common) but using the host with a combination of username and port should be enough.

PR #707, opened 2023-10-20
Update for trilogy

petergoldstein:
Propagate trilogy changes back to database_cleaner.

I'm not entirely sure these methods are used by this gem - they seem ActiveRecord specific - but propagating back (with a minor change, since ActiveRecord is not generally present) for consistency.


etagwerker (approved):
@petergoldstein Looks good, thanks!

PR #719, opened 2025-01-06
Ruby 3.4 quoted locations start with single quotes

pat:
As best I can tell, there's been a change to the output of Ruby's stack traces in `caller` - they now start with a 'normal' single quote, instead of a backtick.

This pattern now handles both, to ensure backwards compatibility.


etagwerker (approved):
@pat Thank you! 💯

PR #725, opened 2026-05-05
Camelize strategy name in shared examples

leoarnold:
Following up on #681, we also need to camelize strategy names in the RSpec shared examples for cleaning strategies.

For developer convenience, the methods for string manipulation were moved from the Cleaner class to a dedicated StringHelper module.


leoarnold:
Remove use of `Enumerable#filter_map` for compatibility with older Rubies.

